PR #34, opened 2020-05-25
Add ability to snooze notifs

Half-Shot:
Partially fixes #34. Still needs support for digest


Half-Shot:
Instead of doing this, we're going to use filters.

PR #47, opened 2021-03-06
Initial work to support discussions

### LINKED ISSUE #43: Support GitHub discussions
Probably fairly important to help moderate the lands. We can wait until there is an actual API though

Half-Shot:
Fixes #43 

Currently looks like:

image


shafqatevo:
Hi, @Half-Shot, is this feature available now in any shape or form? Which client to use and how exactly to set up?

PR #83, opened 2021-11-25
Add provisioning API

Half-Shot:
[Documentation](https://github.com/Half-Shot/matrix-hookshot/blob/hs/provisioning/src/provisioning/api.md)

This PR adds a provisioning HTTP API which allows privileged services to query connection statuses for rooms, connect users via oauth, and various other things.


src/provisioning/api.md
```diff
@@ -0,0 +1,167 @@
+Provisioning API for matrix-hookshot
```
turt2live, line 1: endpoint to get the bot's user ID (for invites/power level) please 😇 

I believe this will be per-service, but happy to be wrong :D

turt2live, line 1: can we get error codes so the provisoner can determine what the failure was without parsing error messages?

Half-Shot: At the moment we use one bot user for all, but I suppose we could start spouting special users for each room.

Half-Shot: Yep, just needs an addition to the ApiError class

Half-Shot: Done!

Half-Shot: The API now exposes a botUserId for all connections. There is also a connectiontypes api to help you search for what's enabled on the bridge.

src/Jira/Router.ts
```diff
@@ -60,3 +62,20 @@ export default class JiraRouter {
...
     }
 }
+
+
+export class JiraProvisionerRouter {
+    constructor(private readonly config: BridgeConfigJira, private readonly tokenStore: UserTokenStore) { }
+
+    public getRouter() {
+        const router = Router();
+        router.get("/oauth", this.onOAuth.bind(this));
```
turt2live, line 72: also need a route to determine if the user is even logged in

src/provisioning/api.md
```diff
@@ -0,0 +1,196 @@
...
+    "id": "opaque-unique-id", // An opaque ID used to refer to this connection. Should **NOT** be assumed to be stable.
+    "service": "jira", // or github, webhook. A human-readable service name to make things look pretty
+    "botUserId": "@hookshot:yourdomain.com", // The bot mxid for the service. Currently this is the sender_localpart, but may change in the future.
+    "config": {
+        // ... connection specific details, can be configured.
+    }
+}
+```
+
+## PUT /v1/{roomId}/connections/{type}
```
turt2live, line 93: is it possible to get an endpoint to list all capable jira projects for the authed user? Ideally in a format that can be picked apart and fed back to this PUT endpoint.

Likewise for github orgs/repos.

Half-Shot: Done in `32ac765` for JIRA, github coming up!

src/Jira/Router.ts
Half-Shot: Done in 32ac765 for JIRA, github coming up!

Half-Shot: GitHub done too now.

src/provisioning/api.md
```diff
@@ -0,0 +1,290 @@
...
+This request is paginated, and `page` sets the page (defaults to `1`) while `perPage` (defaults to `10`) sets the number of entries per page.
+
+This request can be retried until the number of entries is less than the value of `perPage`.
+
+### Response
+
+```json5
+{
+    "loggedIn": true,
+    "repositories": {
```
turt2live, line 206: I think this is supposed to be an array. Likewise for similar endpoints

src/provisioning/api.md
```diff
@@ -0,0 +1,290 @@
...
+Request the status of the users account. This will return a `loggedIn` value to determine if the
+bridge has a JIRA identity stored for the user, and any instances they have access to. Note that if a 
+user does not have access to an instance, they can authenticate again to gain access to it (if they are able
+to consent).
+### Response
+
+```json5
+{
+    "loggedIn": true,
+    "instances": {
```
turt2live, line 268: array?

src/provisioning/api.md
```diff
@@ -0,0 +1,290 @@
...
+
+Request the status of the users account. This will return a `loggedIn` value to determine if the
+bridge has a GitHub identity stored for the user, and any organisations they have access to.
+
+### Response
+
+```json5
+{
+    "loggedIn": true,
+    "organisations": {
```
turt2live, line 186: array?

src/provisioning/api.md
```diff
@@ -0,0 +1,290 @@
...
+        "avatarUrl": "https://avatars.githubusercontent.com/u/8418310?v=4",
+        "description": "A bridge between Matrix and multiple project management services, such as GitHub, GitLab and JIRA. "
+    }
+}
+```
+
+## JIRA
+
+
+### GET /jira/v1/oauth?userId={userId}
```
turt2live, line 242: these endpoints are actually at `/v1/jira`, not `/jira/v1`

likewise for github

src/provisioning/api.md
turt2live: not sure where the resolve button is on this thread, but the thread is resolved - thank you!

src/provisioning/api.md
```diff
@@ -0,0 +1,290 @@
...
+
+```json5
+{
+    "loggedIn": true,
+    "projects": {
+        "key": "PLAY",
+        "name": "Jira Playground",
+        "id": "10015"
+    }
+}
```
turt2live, line 289: This actually returns an array of projects, not an object

src/provisioning/api.md
```diff
@@ -0,0 +1,290 @@
...
+a new JIRA connection.
+### Response
+
+```json5
+{
+    "loggedIn": true,
+    "projects": {
+        "key": "PLAY",
+        "name": "Jira Playground",
+        "id": "10015"
```
turt2live, line 287: `url` instead of `id`

src/provisioning/provisioner.ts
```diff
@@ -0,0 +1,242 @@
...
+        additionalRoutes: {route: string, router: Router}[]) {
+        if (!this.config.secret) {
+            throw Error('Missing secret in provisioning config');
+        }
+        if (!this.config.port) {
+            throw Error('Missing port in provisioning config');
+        }
+        this.expressApp = express();
+        this.expressApp.get("/v1/health", this.getHealth);
+        this.expressApp.use(this.checkAuth.bind(this));
```
turt2live, line 32: need an `this.expressApp.use(bodyParser.json())` here

src/provisioning/provisioner.ts
```diff
@@ -0,0 +1,242 @@
...
+        try {
+            await connection.provisionerUpdateConfig(req.query.userId, req.body);
+        } catch (ex) {
+            next(ex);
+        }
+        res.send(connection.getProvisionerDetails());
+    }
+
+    private async deleteConnection(req: Request<{roomId: string, connectionId: string}>, res: Response<{ok: true}>, next: NextFunction) {
+        const connection = this.connMan.getConnectionById(req.params.roomId, req.params.connectionId);
```
turt2live, line 213: I haven't tried the PATCH or GET endpoints, but this DELETE one doesn't seem to ever find the connection. Tried calling it with `hookshot.getConnections(roomId)[0].id` and it still failed.

I suspect it's a connection ID mismatch or something.

turt2live: on jira btw - probably the same for github?

Half-Shot: I think fixed.

src/provisioning/provisioner.ts
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    private async deleteConnection(req: Request<{roomId: string, connectionId: string}>, res: Response<{ok: true}>, next: NextFunction) {
+        try {
+            const connection = this.connMan.getConnectionById(req.params.roomId, req.params.connectionId);
+            if (!connection) {
+                return next(new ApiError("Connection does not exist", ErrCode.NotFound));
+            }
+            if (!connection.onRemove) {
+                return next(new ApiError("Connection does not support removal", ErrCode.UnsupportedOperation));
```
turt2live, line 220: I believe this is triggering for all integrations. Tested for sure with Jira Projects, but I don't see `onRemove` implemented by anything either.

src/provisioning/api.md
```diff
@@ -0,0 +1,290 @@
...
+### Response
+
+```json5
+{
+    "loggedIn": true,
+    "projects": [
+        "key": "PLAY",
+        "name": "Jira Playground",
+        "url": "https://acme.atlassian.net/projects/PLAY"
+    ]
```
turt2live, line 288: by arrays I meant arrays of objects :p 

applies throughout

src/provisioning/api.md
```diff
@@ -0,0 +1,290 @@
...
+        "url": "https://acme.atlassian.net"
+    ]
+}
+```
+
+### GET /v1/jira/instances/{instanceName}/projects?userId={userId}
+
+Request a list of all projects a user can see in a given instance. The `url` value of a project can be given to create
+a new JIRA connection.
+### Response
```
turt2live, line 279: This actually just returns an unwrapped array, as demonstrated by this helper function:

```
    public async getProjects(instanceName: string): Promise<HookshotJiraProject[]> {
        const bridge = await this.getDefaultBridge();
        return this.doProvisionRequest<HookshotJiraProject[]>(bridge, "GET", `/v1/jira/instances/${instanceName}/projects`);
    }
```

src/Bridge.ts
```diff
@@ -115,6 +121,24 @@ export class Bridge {
...
+                    router: new JiraProvisionerRouter(this.config.jira, this.tokenStore).getRouter(),
+                });
+                this.connectionManager.registerProvisioningConnection(JiraProjectConnection);
+            }
+            if (this.config.github) {
+                routers.push({
+                    route: "/v1/github",
+                    router: new GitHubProvisionerRouter(this.config.github, this.tokenStore).getRouter(),
+                });
+            }
```
turt2live, line 139: looks like this is missing a registration of some kind. I also had to fix the provisioner details of the GithubRepo connection:

```diff
diff --git a/src/Bridge.ts b/src/Bridge.ts
index 2db2d28..ca1e057 100644
--- a/src/Bridge.ts
+++ b/src/Bridge.ts
@@ -136,6 +136,7 @@ export class Bridge {
                     route: "/v1/github",
                     router: new GitHubProvisionerRouter(this.config.github, this.tokenStore).getRouter(),
                 });
+                this.connectionManager.registerProvisioningConnection(GitHubRepoConnection);
             }
             this.provisioningApi = new Provisioner(this.config.provisioning, this.connectionManager, this.as.botIntent, routers);
         }
diff --git a/src/Connections/GithubRepo.ts b/src/Connections/GithubRepo.ts
index 21e7075..7025d5a 100644
--- a/src/Connections/GithubRepo.ts
+++ b/src/Connections/GithubRepo.ts
@@ -682,7 +682,7 @@ ${event.release.body}`;
         return {
             service: "github",
             eventType: GitHubRepoConnection.CanonicalEventType,
-            type: "JiraProject",
+            type: "GithubRepo",
             // TODO: Add ability to configure the bot per connnection type.
             botUserId: botUserId,
         }
```

src/Bridge.ts
```diff
@@ -295,24 +319,23 @@ export class Bridge {
...
-                data: !!(adminRoom),
+                data: !!userId,
                 sender: "Bridge",
                 messageId: msg.messageId,
-                eventName: "response.oauth.response",
+                eventName: "response.github.oauth.response",
             });
         });
 
         this.queue.on<GitHubOAuthTokens>("oauth.tokens", async (msg) => {
```
turt2live, line 332:
```suggestion
        this.queue.on<GitHubOAuthTokens>("github.oauth.tokens", async (msg) => {
```

src/Bridge.ts
```diff
@@ -295,24 +319,23 @@ export class Bridge {
             await this.notifProcessor.onUserEvents(msg.data, adminRoom);
         });
 
-        this.queue.on<OAuthRequest>("oauth.response", async (msg) => {
-            const adminRoom = [...this.adminRooms.values()].find((r) => r.oauthState === msg.data.state);
+        this.queue.on<OAuthRequest>("github.oauth.response", async (msg) => {
+            const userId = this.tokenStore.getUserIdForOAuthState(msg.data.state);
```
turt2live, line 323: to prevent deletion which ends up half-bridging and account:
```suggestion
            const userId = this.tokenStore.getUserIdForOAuthState(msg.data.state, false);
```

src/provisioning/api.md
```diff
@@ -0,0 +1,290 @@
...
+```
+
+# Service specific APIs
+
+Some services have specific APIs for additional functionality, like OAuth.
+
+## GitHub
+
+
+### GET /v1/github/oauth?userId={userId}
```
turt2live, line 162: The URL returned by this works great for authorizing the user, but I think if the github app hasn't been added to the org (including self-org like `turt2live`) then the bridge won't be able to list orgs.

Can we get an endpoint for the second step please? 😇

src/Bridge.ts
```diff
@@ -115,6 +121,24 @@ export class Bridge {
...
+                });
+                this.connectionManager.registerProvisioningConnection(JiraProjectConnection);
+            }
+            if (this.config.github) {
+                routers.push({
+                    route: "/v1/github",
+                    router: new GitHubProvisionerRouter(this.config.github, this.tokenStore).getRouter(),
+                });
+            }
+            this.provisioningApi = new Provisioner(this.config.provisioning, this.connectionManager, this.as.botIntent, routers);
```
turt2live, line 140:
```suggestion
            if (this.config.generic?.enabled) {
                this.connectionManager.registerProvisioningConnection(GenericHookConnection);
            }
            this.provisioningApi = new Provisioner(this.config.provisioning, this.connectionManager, this.as.botIntent, routers);
```

src/Connections/GenericHook.ts
```diff
@@ -119,6 +146,29 @@ export class GenericHookConnection implements IConnection {
...
+    public getProvisionerDetails() {
+        const url = `${this.config.urlPrefix}${this.config.urlPrefix.endsWith('/') ? '' : '/'}${this.hookId}`;
+        return {
+            ...GenericHookConnection.getProvisionerDetails(this.botUserId),
+            id: this.connectionId,
+            config: {
+                transformationFunction: this.transformationFunction,
+                hookId: this.hookId,
+                url,
+            },
```
turt2live, line 168: hook names would be great, if only for display purposes 😇 

I think this should be relatively straightforward, but a bit larger than a github suggestion can handle

src/provisioning/api.md
```diff
@@ -0,0 +1,290 @@
...
+{
+    // ... connection specific details, can be configured.
+}
+```
+### Response
+
+```json5
+{
+    // The eventId of the state event that describes the connection.
+    "eventId": "!abc:def"
```
turt2live, line 112: echoing the connection back to the caller is a bit more useful than the event ID. With an event ID, the caller needs to have context of the room which might not be possible (particularly in the case of integration managers) whereas the connection information means that the caller can simply append the response to a local array instead of having to re-fetch all connections again.

Half-Shot: This was a bit difficult due to our asynchronous processing of new connections, but I have hopefully solved this now.

src/Github/GithubInstance.ts
```diff
@@ -122,6 +132,11 @@ export class GithubInstance {
     public onInstallationRemoved(data: GitHubWebhookTypes.InstallationDeletedEvent|GitHubWebhookTypes.InstallationSuspendEvent) {
         this.installationsCache.delete(data.installation.id);
     }
+
+    public get newInstallationUrl() {
+        // E.g. https://github.com/apps/matrix-bridge/installations/new
+        return `https://github.com/apps/${this.appName}/installations/new`;
```
turt2live, line 138: the URL uses the slug, not the name.

```ts
const appDetails = await this.internalOctokit.apps.getAuthenticated();
this.internalAppName = appDetails.data.name;
this.internalAppSlug = appDetails.data.slug;
```

src/provisioning/api.md
```diff
@@ -0,0 +1,292 @@
...
+
+### Response
+
+```json5
+{
+    "loggedIn": true,
+    "organisations":[{
+        "name": "half-shot",
+        "avatarUrl": "https://avatars.githubusercontent.com/u/8418310?v=4"
+    }]
```
turt2live, line 194: not really sure what's going on with the two oauth URLs, but I think the bridge needs to probably request more oauth permissions for the user URL and then when returning organizations/repos filter out the ones that the org auth can't handle. 

UX would be something like:
```
Before we can bridge your repositories, please allow the bridge to interact with github as you:

[Authorize]
```

```
[Click here] to add the bridge to your organization. 

Organization: [ > list] [Refresh]
Repository: [ > list]
[Bridge]
```

src/Github/Router.ts
```diff
@@ -0,0 +1,153 @@
...
+                });
+            }
+        
+            return res.send({
+                page,
+                repositories,
+            });
+        } catch (ex) {
+            log.warn(`Failed to fetch accessible repos for ${req.params.orgName} / ${req.query.userId}`, ex);
+            return next(new ApiError("Could not fetch accessible repos for GitHub org", ErrCode.Unknown));
```
turt2live, line 111:
```suggestion
            return next(new ApiError("Could not fetch accessible repos for GitHub org", ErrCode.AdditionalActionRequired));
```

src/Github/Router.ts
```diff
@@ -0,0 +1,153 @@
...
+            }
+        
+            return res.send({
+                page,
+                repositories,
+                ...(orgRes.data.repository_selection === 'selected' && {changeSelectionUrl: `https://github.com/settings/installations/${userInstallation.data.id}`})
+            });
+        } catch (ex) {
+            log.warn(`Failed to fetch accessible repos for ${req.query.userId}`, ex);
+            return next(new ApiError("Could not fetch accessible repos for GitHub user", ErrCode.Unknown));
```
turt2live, line 150:
```suggestion
            return next(new ApiError("Could not fetch accessible repos for GitHub user", ErrCode.AdditionalActionRequired));
```

src/Github/Router.ts
```diff
@@ -0,0 +1,153 @@
...
+            const orgRes = await octokit.orgs.listForAuthenticatedUser({page, per_page: perPage});
+            for (const org of orgRes.data) {
+                organisations.push({
+                    name: org.login,
+                    avatarUrl: org.avatar_url,
+                });
+            }
+        } catch (ex) {
+            log.warn(`Failed to fetch orgs for GitHub user ${req.query.userId}`, ex);
+            return next( new ApiError("Could not fetch orgs for GitHub user", ErrCode.Unknown));
```
turt2live, line 74:
```suggestion
            return next( new ApiError("Could not fetch orgs for GitHub user", ErrCode.AdditionalActionRequired));
```

turt2live (commented):
docs & improved error codes so integration managers can do UI with the errors

src/provisioning/api.md
turt2live: For clarity: I've finally narrowed this down to my usage of the org-specific and account-specific endpoints. Ignoring `organisations` and the `/orgs/:orgId/repositories` endpoint leaves just `/repositories` which works as expected. 

I did also have some switching logic to handle the two now-unused-by-me endpoints, but as implied I don't really need it anymore. It was to handle the special case of the user's own account not being counted as an org:
image
image

turt2live (commented):
I ended up doing some fixes on my own branch to match the integration manager experience: https://github.com/Half-Shot/matrix-hookshot/pull/91

turt2live (approved):
lgtm with https://github.com/Half-Shot/matrix-hookshot/pull/83#discussion_r763423677 fwiw :)

PR #105, opened 2021-12-23
Figma Support

### LINKED ISSUE #100: Add support for figma
Essentially replace https://github.com/Half-Shot/matrix-figma

Half-Shot:
Fixes #100 

This is based off my previous bridge, https://github.com/Half-Shot/matrix-figma. I'll likely deprecate that bridge once this lands.


docs/setup/figma.md
```diff
@@ -0,0 +1,42 @@
...
+ - A figma account that is on the professional tier, as the free tier does provide webhook access.
+ - Your team ID. You can get this by going to the team page on Figma, and looking for the ID in the url (e.g. 12345 in `https://www.figma.com/files/team/12345/...`)
+
+## Configuration
+
+You can now set some configuration in the bridge `config.yml`
+
+```yaml
+figma:
+  publicUrl: https://example.com/hookshot/
```
HarHarLinks, line 16: @Half-Shot this doesn't seem right according to the below description. should be `https://example.com/hookshot/figma/webhook`?

HarHarLinks: It also feels worth pointing out the inconsistence that the other services do not appear to have such specific `/service/webhook` endpoints

Half-Shot: Ah, the below description is wrong, will fix.

PR #115, opened 2021-12-31
Update setup bot command for webhooks

puffnfresh:


docs/setup/webhooks.md
```diff
@@ -29,7 +29,7 @@ webhook requests from `https://example.com/mywebhookspath` to the bridge, an exa
 To add a webhook to your room:
   - Invite the bot user to the room.
   - Make sure the bot able to send state events (usually the Moderator power level in clients)
-  - Say `!setup webhook`
+  - Say `!hookshot webhook example`
```
Half-Shot, line 32: Small detail.

```suggestion
  - Say `!hookshot webhook example` where `example` is a name for your hook.
```

Half-Shot (commented):
Otherwise looks good to me!

Could you add a file to https://github.com/puffnfresh/matrix-hookshot/tree/allow-generic-only/changelog.d called `115.doc` that briefly explains your change in a line?

PR #124, opened 2022-01-04
Docs and config improvements

### LINKED ISSUE #120: clarify resource listeners
https://github.com/Half-Shot/matrix-hookshot/blob/e6dd83ac5bd57a9e189b74999879efb767331646/docs/setup.md#L68

some of these are implicitly clear to me, but others are not
- webhooks: the part that received webhooks
- widgets: ?
- metrics: prometheus metrics
- provisioning: provisioning api
- appservice: ? maybe the same as`bridge.bindAddress` and`bridge.port`?

additionally, please explain how these relate to keys such as `bridge.bindAddress` and `bridge.port`

### LINKED ISSUE #123: generic webhooks + markdown, and docs
if I receive a "generic" webhook without jstransform stuff, would it be a plain m.text without formatting? it says that markdown is rendered after jstransform if you use them.
> If the body contains a `text` key, then that key will be used as a message body in Matrix.
> result will be automatically transformed by a Markdown parser.

I would find it very useful to send json with a `formatted_text` key which would be used literally as `formatted_body` of a org.matrix.custom.html message. is markdown applied to `text`?
> If the body also contains a `username` key, then the message will be prepended by the given username.

so, like a relaybot bridge? would be nice to have "puppeting" this way, as matrix-appservice-webhooks used to

Half-Shot:
Fixes #120 
Fixes #123

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -63,6 +63,7 @@ on how to add appservices. [Synapse documents the process here](https://matrix-o
 You will need to configure some listeners to make the bridge functional.
 
 ```yaml
+listeners:
   # (Optional) HTTP Listener configuration.
   # Bind resource endpoints to ports and addresses.
   # 'resources' may be any of webhooks, widgets, metrics, provisioning
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -65,13 +65,12 @@ You will need to configure some listeners to make the bridge functional.
 ```yaml
   # (Optional) HTTP Listener configuration.
   # Bind resource endpoints to ports and addresses.
-  # 'resources' may be any of webhooks, widgets, metrics, provisioning, appservice
+  # 'resources' may be any of webhooks, widgets, metrics, provisioning
   #
   - port: 9000
     bindAddress: 0.0.0.0
     resources:
       - webhooks
-      - widgets
   - port: 9001
     bindAddress: 127.0.0.1
     resources:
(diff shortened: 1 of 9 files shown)


docs/setup.md
```diff
@@ -80,12 +80,41 @@ You will need to configure some listeners to make the bridge functional.
...
-port, or one on each.
+port, or one on each. Each listener MUST listen on a unique port.
 
 You will also need to make this port accessible to the internet so services like GitHub can reach the bridge. It
 is recommended to factor hookshot into your load balancer configuration, but currrently this process is left as an
 excercise to the user.
 
+In terms of API endpoints:
+
+- The `webhooks` resource handles resources under `/`, so it should be on it's own listener.
```
HarHarLinks, line 91: > should be on its on listener

but you can reasonably assume that webhooks will not get assigned an endpoint `/metrics` or `/v1...`.... right?!

HarHarLinks, line 91: *its

docs/setup.md
```diff
@@ -80,12 +80,41 @@ You will need to configure some listeners to make the bridge functional.
...
+  # Basic homeserver configuration
+  #
+  domain: example.com
+  url: http://localhost:8008
+  mediaUrl: http://example.com
+  port: 9993
+  bindAddress: 127.0.0.1
+```
+
+The `port` and `bindAddress` must not conflict with the other listeners in the bridge config. This listeners should **not** be reachable
```
HarHarLinks, line 114: *listener

docs/setup.md
```diff
@@ -80,12 +80,41 @@ You will need to configure some listeners to make the bridge functional.
...
+</section>
+
+
+```yaml
+bridge:
+  # Basic homeserver configuration
+  #
+  domain: example.com
+  url: http://localhost:8008
+  mediaUrl: http://example.com
```
HarHarLinks, line 109: perhaps it would be reasonable to explain:
- domain: "name" of your homeserver, as it appears in your MXIDs
- url: where hs is reachable, preferably internally i.e. localhost => docker needs adjustment
- mediaUrl: homeserver, but via potential load balancer etc to reach the media worker(?)
- port & bindAddress need adjustment/forwarding when using docker

HarHarLinks (commented):
some comments

HarHarLinks:
@Half-Shot not sure if this shows up in your notifs, hence ping: please look over these comments

PR #128, opened 2022-01-05
Support listening for incoming mails over IMAP

Half-Shot:
This is a barebones PR to listen over IMAP for mails and then notify them in the notification room.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,10 +12,28 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


Half-Shot:
Hey there people on the tail end of 2025, closing this PR as we've in practice not felt the need to work on it over the last few years and it's now rotted to the point where it would be a lot of work to fix. Anyone may pick this back up if they wish to, however!

PR #167, opened 2022-01-12
Add support for a tiered permission system

### LINKED ISSUE #74: Add tiered bridge permissions to limit who can use the bridge
Similar to the mautrix set of bridges / IRC bridges, we'd like to allow some people to read only, some to set a token to perform actions and some should be able to read notifications.

Half-Shot:
Fixes #74 

This PR allows the bridge admin to control who can use the bridge and at what level. This tries to mimick the mautrix system of having actor names, and the level of service they can access. Hookshot additionally allows you to control which services the users can access on top of functionality.

This PR also changes the default permissions, as before we had no permission system:

**ALL** users who are on the same homeserver as the bridge will have `admin` level control, whereas everyone else will have no access. This may not be suitable for your deployment, and you should set your own permission settings before restarting.


src/ConnectionManager.ts
```diff
@@ -85,6 +88,9 @@ export class ConnectionManager {
             if (!this.config.github || !this.config.github.oauth || !this.github) {
                 throw Error('GitHub is not configured');
             }
+            if (!this.config.checkPermission(userId, "github", BridgePermissionLevel.manageConnections)) {
+                throw new ApiError('User is not permitted to provision connections for Jira', ErrCode.ForbiddenUser);
```
HarHarLinks, line 92: this should be github

config.sample.yml
```diff
@@ -91,6 +91,13 @@ logging:
   # (Optional) Logging settings. You can have a severity debug,info,warn,error
   #
   level: info
+permissions:
+  # (Optional) Permissions for using the bridge
+  #
+  - actor: example.com
```
HarHarLinks, line 97: is "actor" a (regex?) value for end user's mxid?

config.sample.yml
```diff
@@ -91,6 +91,13 @@ logging:
...
+    services:
+      - service: "*"
+        level: admin
```
HarHarLinks, line 100: need to document possible values and meanings

HarHarLinks (changes requested):
Don't take this review as much more than a rough sanity and spelling check, I didn't get into any functional stuff.

Went in here originally looking for docs, but there are none yet.

config.sample.yml
Half-Shot: I've hopefully documented this on the docs site now.

config.sample.yml
Half-Shot: See setup.md

PR #184, opened 2022-01-31
Do not explicitly pass --target to rustup

spantaleev:
It doesn't seem to be necessary to install Rust.
Passing `--target` prevents the Dockerfile from being built on different architectures.
Not passing it makes it auto-detect what to install.

Actually, I've managed to build this Dockerfile on ARM64 even with the hardcoded `--target`. It's misleading though, so removing it is better.

Discussed here: https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/1505

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ COPY . /src
 WORKDIR /src
 
 # We need rustup so we have a sensible rust version, the version packed with bullsye is too old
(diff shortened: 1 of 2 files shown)


Half-Shot:
I can't remember why we did this initially, it might well have been my fruitless efforts to use Rust with alpine.

Half-Shot:
@spantaleev Can I have a synapse-style changelog entry just describing what this does?

spantaleev:
I've added a changelog entry. I hope it's all good now!

PR #189, opened 2022-02-13
fix #162

### LINKED ISSUE #162: registration needs user pattern that allows generic.userIdPrefix
https://github.com/Half-Shot/matrix-hookshot/blob/f56105db9de8a8ed2ef4597c28d4e8d799a2cccd/registration.sample.yml#L7

https://github.com/Half-Shot/matrix-hookshot/blob/f56105db9de8a8ed2ef4597c28d4e8d799a2cccd/config.sample.yml#L56

I only have
```yaml
  users:
    - regex: "@_github_.*:{{ matrix_domain }}"
      exclusive: true
  aliases:
    - regex: "#github_.+:{{ matrix_domain }}"
      exclusive: true
```
but
```yaml
userIdPrefix: webhooks_
```

both as provided by the docs, and am getting error `Invalid user localpart for this application service.` on `http://matrix-nginx-proxy:12080/_matrix/client/r0/register?user_id=%40webhooks_...`, so may assume this is the reason?

Perhaps this is the time to mention that appservice-webhooks uses `_webhook` as prefix, so admins should take care these don't clash https://github.com/turt2live/matrix-appservice-webhooks/blob/a2ef5051fa8e7ba62b3a0035452bc76c9f6dc1f1/config/sample.yaml#L13

HarHarLinks:
maybe?

### DIFF
--- a/registration.sample.yml
+++ b/registration.sample.yml
@@ -3,11 +3,17 @@ as_token: aaaaaaaaaa # This again can be a random string
 hs_token: aaaaaaaaaa # ..as can this
 namespaces:
   rooms: []
-  users:
-    - regex: "@_github_.*:foobar" # Where foobar is your homeserver's domain
+  users: # In the following, foobar is your homeserver's domain
+    - regex: "@_github_.*:foobar"
+      exclusive: true
+    - regex: "@_gitlab_.*:foobar"
+      exclusive: true
+    - regex: "@_jira_.*:foobar"
+      exclusive: true
+    - regex: "@_webhooks_.*:foobar" # Where _webhooks_ is set by userIdPrefix in config.yml
       exclusive: true
   aliases:
-    - regex: "#github_.+:foobar"  # Where foobar is your homeserver's domain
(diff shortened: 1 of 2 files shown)


registration.sample.yml
```diff
@@ -3,11 +3,17 @@ as_token: aaaaaaaaaa # This again can be a random string
 hs_token: aaaaaaaaaa # ..as can this
```
Half-Shot, line 3: Certainly better, can we get a changelog? :)

HarHarLinks: would that be changelog.d/162.doc? I'm not sure what format you follow, could you add a CONTRIBUTING.md file linking to towncrier or https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog or so?

Half-Shot: it's in progress https://github.com/Half-Shot/matrix-hookshot/pull/134/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R56

PR #227, opened 2022-03-07
Move webhooks from / to /webhook

Half-Shot:
The generic webhooks router currently listens on `/:hookId/` which means that while the request paths are short, it causes confusion when a user misconfigures their bridge and gets a 200 response regardless. This PR makes the bridge *prefer* `/webhook`, while still supporting `/:hookId`. 

In addition, the bridge now gives an appropriate 404 from both paths when a hookId is not found. The bridge *can* also be configured to wait until the webhook has finished processing before responding, although this behavior is off by default.

### DIFF
--- a/changelog.d/227.feature
+++ b/changelog.d/227.feature
@@ -1,0 +1,1 @@
+Generic webhooks now listen for incoming hooks on `/webhooks`. Existing setups using `/` will continue to work, but should be migrated where possible. See [the documentation](https://matrix-org.github.io/matrix-hookshot/setup/webhooks.html#configuration) for more information.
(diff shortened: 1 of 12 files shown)


HarHarLinks:
can you create an issue and ping me on https://github.com/spantaleev/matrix-docker-ansible-deploy when you merge?

Half-Shot:
> can you create an issue and ping me on https://github.com/spantaleev/matrix-docker-ansible-deploy when you merge?

Done - https://github.com/spantaleev/matrix-docker-ansible-deploy/issues/1681

PR #245, opened 2022-03-26
Update github_repo.md

HarHarLinks:
- really the first step should be to ensure that the app is installed before you even try to connect it
- clarify creation time room settings
- "moderator permissions" isn't really accurate, just default or convention (perhaps even element convention rather than matrix convention?)
- typos

### DIFF
--- a/docs/usage/room_configuration/github_repo.md
+++ b/docs/usage/room_configuration/github_repo.md
@@ -10,15 +10,16 @@ a new release.
 
 To set up a connection to a GitHub Repository in a new room:
 
-(N.B you must have permission to bridge GitHub repositories before you can use this command, see [auth](../auth.html#github))
+(NB you must have permission to bridge GitHub repositories before you can use this command, see [auth](../auth.html#github).)
 
-1. Invite the bridge bot (e..g `@hookshot:example.com`)
-2. Give the bridge bot moderator permissions or higher (power level 50).
-3. Send the command `!hookshot github repo https://github.com/my/project`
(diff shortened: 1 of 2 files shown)


HarHarLinks:
See also #260

docs/usage/room_configuration/github_repo.md
```diff
@@ -10,15 +10,16 @@ a new release.
...
-1. Invite the bridge bot (e..g `@hookshot:example.com`)
-2. Give the bridge bot moderator permissions or higher (power level 50).
-3. Send the command `!hookshot github repo https://github.com/my/project`
-4. If you have permission to bridge this repo, the bridge will respond with a confirmation message.
-5. Note: The bridge will need to either:
-    - Have a GitHub installation registered with the organisation
+(NB you must have permission to bridge GitHub repositories before you can use this command, see [auth](../auth.html#github).)
+
+1. The bridge will need to either:
+    - Have a GitHub installation registered with the organisation (can be user)
```
Half-Shot, line 16: "user" just feels a bit vague here

```suggestion
    - Have a GitHub installation registered with the organisation (or GitHub user account)
```

Half-Shot (approved):
Otherwise good!

HarHarLinks:
> Otherwise good!

I haven't looked to closely whether there is more in #260 that should be addressed. You should be able to commit to this PR (my forked branch)

Half-Shot:
I think it's fine to go incremental in these cases :). We can fix #260 in subsequent PRs.

PR #273, opened 2022-04-03
Made ESLint lint all TypeScript files and fix a few linter errors

jaller94:
* Lint all files by quoting the paths in `package.json`. Currently, `*.ts` files directly in the `src` folder are not linted on all operating systems because `**` resolves to folder names.
* Lint the `test` and `web ` folder with slightly adapted rules.
* Fix all ESLint errors
* Fix a few ESLint warnings.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,20 +2,55 @@ module.exports = {
     root: true,
     parser: '@typescript-eslint/parser',
     plugins: [
-        '@typescript-eslint',
-        'mocha'
+        '@typescript-eslint'
     ],
     extends: [
-        'eslint:recommended',
         'plugin:@typescript-eslint/recommended',
     ],
+    // eslint-config-preact needs a Jest version to be happy, even if Jest isn't used.
+    // See https://github.com/preactjs/eslint-config-preact/issues/19#issuecomment-997924892
+    settings: {
+        jest: { "version": 27 },
+    },
     rules: {
         "@typescript-eslint/explicit-module-boundary-types": "off",
         "camelcase": ["error", { "properties": "never", "ignoreDestructuring": true }],
         "no-console": "error"
     },
-    "env": {
-        "node": true,
-        "es6": true,
+    env: {
+        node: true,
+        es6: true,
     },
+    overrides: [
+        {
+            files: ["test/**/*.ts"],
+            parser: '@typescript-eslint/parser',
+            plugins: [
+                '@typescript-eslint',
+                'mocha',
+            ],
+        },
+        {
+            files: ["web/**/*.ts", "web/**/*.tsx"],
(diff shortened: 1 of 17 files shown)


Half-Shot (approved):

Half-Shot (changes requested):
Ah sorry, we **must** include the `preact` imports in preact code or the compilation will fail. Eslint doesn't know that it's needed.

There is possibly a linter rule exemltion we need to import for this, I've not checked

package.json
```diff
@@ -28,7 +28,7 @@
     "start:matrixsender": "node --require source-map-support/register lib/App/MatrixSenderApp.js",
     "test": "mocha -r ts-node/register tests/init.ts tests/*.ts  tests/**/*.ts",
     "lint": "yarn run lint:js && yarn run lint:rs",
-    "lint:js": "eslint -c .eslintrc.js src/**/*.ts",
+    "lint:js": "eslint -c .eslintrc.js 'src/**/*.ts' 'tests/**/*.ts' 'web/**/*.ts' 'web/**/*.tsx'",
```
jaller94, line 31: For `**` to work reliably the string has to be quoted. Otherwise, it might get resolved by the operating system.

.eslintrc.js
```diff
@@ -2,20 +2,55 @@ module.exports = {
...
+            ],
+            plugins: [
+                '@typescript-eslint',
+            ],
+            rules: {
+                "no-console": "off",
+                "no-unused-vars": "off",
+                "@typescript-eslint/no-unused-vars": ["error"],
+                "no-useless-constructor": "off",
+                "@typescript-eslint/no-useless-constructor": ["error"],
```
jaller94, line 52: I couldn't figure out why the regular eslint rules were still being used.
Only the TypeScript variants make sense for this code base.

.eslintrc.js
```diff
@@ -2,20 +2,56 @@ module.exports = {
...
     plugins: [
-        '@typescript-eslint',
-        'mocha'
+        '@typescript-eslint'
     ],
     extends: [
-        'eslint:recommended',
-        'plugin:@typescript-eslint/recommended',
+        "eslint:recommended",
+        "plugin:@typescript-eslint/recommended",
```
Half-Shot, line 9:
```suggestion
        'eslint:recommended',
        'plugin:@typescript-eslint/recommended',
```

So that the diff is a touch neater

.eslintrc.js
Half-Shot: Hm, suppose we can come back to this one when it poses a problem for us. The linter seems happy for the moment.

Half-Shot (approved):
Looks good, thanks a ton for cleaning up the .eslintrc.js file :)

Issue #179, opened 2022-01-18
add ability to format generic webhooks as notice, emote


HarHarLinks:
The old appservice could send emote and more importantly notice msgtypes: https://github.com/turt2live/matrix-appservice-webhooks#json-body-for-posting-messages

Hookshot should offer the same functionality to support this. It should be settable by a `msgtype` field in the webhook data or similar, and ideally also via the jstransform option. This would enable some pretty [neat](https://github.com/HarHarLinks/matrix-webhook-receiver#noisy-notifications) use cases.

Half-Shot:
This is fixed via transform functions in https://github.com/matrix-org/matrix-hookshot/pull/282.

PR #282, opened 2022-04-09
Generic webhook transform: allow specifying msgtype

SpiritCroc:
So we can send noisy m.text notifications.

Addresses the jstransform part of https://github.com/matrix-org/matrix-hookshot/issues/179.

Signed-off-by: Tobias Büttner <dev@spiritcroc.de>

### DIFF
--- a/changelog.d/282.feature
+++ b/changelog.d/282.feature
@@ -1,0 +1,1 @@
+Allow specifying msgtype for generic webhook transformations.
--- a/docs/setup/webhooks.md
+++ b/docs/setup/webhooks.md
@@ -109,6 +109,7 @@ The `v2` api expects an object to be returned from the `result` variable.
   "empty": true|false, // Should the webhook be ignored and no output returned. The default is false (plain must be provided).
   "plain": "Some text", // The plaintext value to be used for the Matrix message.
   "html": "<b>Some</b> text", // The HTML value to be used for the Matrix message. If not provided, plain will be interpreted as markdown.
+  "msgtype": "some.type", // The message type, such as m.notice or m.text, to be used for the Matrix message. If not provided, m.notice will be used.
 }
 ```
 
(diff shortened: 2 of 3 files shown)


Half-Shot (approved):
This looks great! No concerns here. Can you:

1. Sign off as per https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off (sorry, need to include those instructions in this repo)
2. Add a changelog entry https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog
3. Add a note in https://github.com/matrix-org/matrix-hookshot/blob/main/docs/setup/webhooks.md to document the new feature.

Once that's done, we can merge

SpiritCroc:
Thanks, updated!

SpiritCroc:
Just noticed this probably doesn't fully fix https://github.com/matrix-org/matrix-hookshot/issues/179, since this only applies for transformations and not for webhooks without transformation js, I think

Half-Shot:
>Just noticed this probably doesn't fully fix https://github.com/matrix-org/matrix-hookshot/issues/179, since this only applies for transformations and not for webhooks without transformation js, I think

Up to you on this one. I think the transformation functions probably solve it on a technicality, but if you like you could add a config option to the state event (https://github.com/matrix-org/matrix-hookshot/pull/282/files#diff-41e6a8f8490f260df3e16fd6864d0defb8c7e2eaa7cd1627ce25bd681b90ff40R22) to set the msgtype

SpiritCroc:
The current PR is all I need for my use case right now. I just removed the "Fixes" keyword from the PR message, since #179 is technically not fully fixed with this, I guess :)

Half-Shot:
I'd actually say it does cover the requested feature in the issue, and that we can consider a non-transformation way an extension task for someone to fix if they don't want to use transformations :+1:

PR #309, opened 2022-04-16
Multiple improvements to GitLab push event formatting

HarHarLinks:
This PR accomplishes ~2~ these changes to the rendering of GitLab events:

- [x] commit hashes formatted as inline code
- [x] "N commits" links to the list of commits on that branch instead of just the last commit individually
- [x] only show list of commits if it is complete
- [x] list commiter names unless all commits were authored and pushed by the same person (identified by email)
- [x] open issue about configuring max length of commits list
- [x] open issue for configuring `<details>` commit lists

I also want to discuss:
- Why format the branch name as inline code?
- You limit the amount of commits to the last 5 in case of pushing more that that at once. I can see the reasoning behind that, you might not want to spam the room with loads of commits when you subscribe to pushes in order to be notified that there was a push. As reader you can infer that some are hidden by the total number of commits. Alternatively you could only include the list if it is "complete" and not include a list but only the link to "N commits" when it wouldn't be.
- On top of this, GitLab limits itself to including at most 20 commits in a webhook anyway. I would like the bridge to remind me in these cases that while it says "pushed 20 commits" there may actually be more, i.e. say "pushed 20+ commits" or similar. 
- Personally my workflow using GitLab webhooks (with the old bridge and [some jinja](https://github.com/HarHarLinks/matrix-webhook-receiver/blob/main/examples/gitlab.jinja2)) has been to review all of the commit messages on matrix, because after all I want to be lazy and _not_ have to click these links. :smile: How hard would it be to make the limit of 5 commits configurable?
- Can't you have multiple gitlab hooks per room? If you can, then perhaps the repo should be more prominent, e.g. `**strong**`, hyperlinked and in the first line before everything else, like a heading (but an actual `<h1>` would be too much)

Signed-off-by: Kim Brose <kim.brose@rwth-aachen.de>


Half-Shot:
Sorry, pressed the button without thinking

HarHarLinks:
> Sorry, pressed the button without thinking

Yeah, it is "Ready for review" in the sense that I want you to look at it, I drafted because I wanted to discuss before merging.

Half-Shot:
>Why format the branch name as inline code?

I copied it from the tag push code I suspect. I don't really have an opinion on keeping it that way.

>You limit the amount of commits to the last 5 in case of pushing more that that at once. I can see the reasoning behind that, you might not want to spam the room with loads of commits when you subscribe to pushes in order to be notified that there was a push. As reader you can infer that some are hidden by the total number of commits. Alternatively you could only include the list if it is "complete" and not include a list but only the link to "N commits" when it wouldn't be.

I would be okay with doing that, but maybe include the last commit so there is "some" context about what happened (e.g. a merge, fix, whatever). Possibly the same behaviour as with 1 commit, but with a link to the rest.

>On top of this, GitLab limits itself to including at most 20 commits in a webhook anyway. I would like the bridge to remind me in these cases that while it says "pushed 20 commits" there may actually be more, i.e. say "pushed 20+ commits" or similar.

GitLab should still give you the total commit count (in `total_commits_count`), so we should probably use that when doing the above behavior.

>Personally my workflow using GitLab webhooks (with the old bridge and [some jinja](https://github.com/HarHarLinks/matrix-webhook-receiver/blob/main/examples/gitlab.jinja2)) has been to review all of the commit messages on matrix, because after all I want to be lazy and not have to click these links. smile How hard would it be to make the limit of 5 commits configurable?

Happy for you to make that configurable via room state :)

>Can't you have multiple gitlab hooks per room? If you can, then perhaps the repo should be more prominent, e.g. **strong**, hyperlinked and in the first line before everything else, like a heading (but an actual <h1> would be too much)

I think there might be some prior art in the GitHubRepo code you could use as inspiration?

image

Having it linked felt prominent enough to me.

HarHarLinks:
> > You limit the amount of commits to the last 5 in case of pushing more that that at once. I can see the reasoning behind that, you might not want to spam the room with loads of commits when you subscribe to pushes in order to be notified that there was a push. As reader you can infer that some are hidden by the total number of commits. Alternatively you could only include the list if it is "complete" and not include a list but only the link to "N commits" when it wouldn't be.
> 
> I would be okay with doing that, but maybe include the last commit so there is "some" context about what happened (e.g. a merge, fix, whatever). Possibly the same behaviour as with 1 commit, but with a link to the rest.

This sounds good

> > On top of this, GitLab limits itself to including at most 20 commits in a webhook anyway. I would like the bridge to remind me in these cases that while it says "pushed 20 commits" there may actually be more, i.e. say "pushed 20+ commits" or similar.
> 
> GitLab should still give you the total commit count (in `total_commits_count`), so we should probably use that when doing the above behavior.

I somehow missed this field existed, excellent.

> > Personally my workflow using GitLab webhooks (with the old bridge and [some jinja](https://github.com/HarHarLinks/matrix-webhook-receiver/blob/main/examples/gitlab.jinja2)) has been to review all of the commit messages on matrix, because after all I want to be lazy and not have to click these links. smile How hard would it be to make the limit of 5 commits configurable?
> 
> Happy for you to make that configurable via room state :)

Oof ok I declare that beyond the scope of this quick PR but maybe eventually when it's rusted so I can tackle that and learn rust instead of js :stuck_out_tongue_winking_eye: 

> > Can't you have multiple gitlab hooks per room? If you can, then perhaps the repo should be more prominent, e.g. **strong**, hyperlinked and in the first line before everything else, like a heading (but an actual \#  would be too much)
> 
> I think there might be some prior art in the GitHubRepo code you could use as inspiration?
> 
> image
> 
> Having it linked felt prominent enough to me.

I don't think I understand what you're saying. Anyway here is a comparison of a GitLab push using
1. [my jinja](https://github.com/HarHarLinks/matrix-webhook-receiver/blob/main/examples/gitlab.jinja2). Somehow this order makes more sense to me, but it may or not have been based on 3.
image

2. current hookshot i.e. before this PR
image

3. Whatever nheko uses (go-neb? @deepbluev7)
image

HarHarLinks:
Re: lists of commits, I've seen https://matrix.to/#/#android-announcements:schildi.chat use `<details>` (which is supported by matrix spec, but far from implemented in all clients)

image

src/Connections/GitlabRepo.ts
```diff
@@ -222,12 +222,16 @@ export class GitLabRepoConnection extends CommandConnection {
             return;
         }
         const branchname = event.ref.replace("refs/heads/", "");
-        const commitsurl = `${event.project.homepage}/-/commits/${branchname}`
+        const commitsurl = `${event.project.homepage}/-/commits/${branchname}`;
         const branchurl = `${event.project.homepage}/-/tree/${branchname}`;
         const shouldName = !event.commits.every(c => c.author.name === event.commits[0]?.author.name);
 
+        const maxCommits = 5;
```
Half-Shot, line 229: if this is a const, just define it at the top of the file as `PUSH_MAX_COMMITS` or something.

Half-Shot (approved):
Seems reasonable otherwise.

Half-Shot:
@HarHarLinks can you make the title more helpful too :)

changelog.d/309.feature
```diff
@@ -0,0 +1 @@
+Improve GitLab push hook formatting
```
Half-Shot, line 1: Can you expand on this, something about only showing one commit if this goes over the limit.

src/Connections/GitlabRepo.ts
HarHarLinks: it's just a helper for now, it should eventually be read from room state #329

src/Connections/GitlabRepo.ts
```diff
@@ -222,13 +222,17 @@ export class GitLabRepoConnection extends CommandConnection {
             return;
         }
         const branchname = event.ref.replace("refs/heads/", "");
-        const commitsurl = `${event.project.homepage}/-/commit/${event.after}`
+        const commitsurl = `${event.project.homepage}/-/commits/${branchname}`;
         const branchurl = `${event.project.homepage}/-/tree/${branchname}`;
         const shouldName = !event.commits.every(c => c.author.name === event.commits[0]?.author.name);
```
HarHarLinks, line 227: @Half-Shot: I think this is wrong, it should rather check if the pusher and commiter are the same.
The way it currently is, I can pull a commit from upstream and push it to my forked remote and it will only tell me that I pushed it but not the original commiter, who is someone else (actually happened today).

src/Connections/GitlabRepo.ts
Half-Shot: Yeah we'd likely still want a default.

src/Connections/GitlabRepo.ts
Half-Shot: Sure, I think that's reasonable. Perhaps just change the if statement to check against the pusher email address? (since that's the best unique thing we have)

src/Connections/GitlabRepo.ts
```diff
@@ -239,13 +239,13 @@ export class GitLabRepoConnection extends CommandConnection {
...
             content += `\n - ${commits}\n`;
-        } else if (event.commits.length === 1) {
-            content += ` - ${commits}`;
+        } else if (displayedCommits === 1) {
+            content += `: ${commits}`;
         }
         if (tooManyCommits) {
-            content += `. View [${event.total_commits_count - 1} more](${commitsurl}).`;
+            content += `, and [${event.total_commits_count - 1} more](${commitsurl}) commits`;
         }
```
HarHarLinks, line 441: This behaviour could be moved inside `else if (displayedCommits === 1)` as it is only triggered then.
What do you think, readability-wise?

src/Connections/GitlabRepo.ts
HarHarLinks: so, technically shouldName could also depend on tooManyCommits, but i'm not sure it's worth doing

PR #314, opened 2022-04-21
Support notifying on gitlab review comments

Half-Shot:
This one was a bit tricky. GitLab doesn't have a notion of "review comments" so we've had to play a bit of a game to make it work like we have on GitHub.

Specifically, we determine a GitLab review by the following:
- Did a note come in for a Merge Request
- Was the note *not* from the original MR author

We also batch together several notes for the same MR into one matrix message if several of them arrive in a short period of time (5s), since GitLab sends the review comments at the same time.

I'm not particularly happy with this approach, but I think it's probably better than not having them.

### DIFF
--- a/src/Connections/GitlabRepo.ts
+++ b/src/Connections/GitlabRepo.ts
@@ -7,7 +7,7 @@ import { MatrixEvent, MatrixMessageContent } from "../MatrixEvent";
 import markdown from "markdown-it";
 import LogWrapper from "../LogWrapper";
 import { GitLabInstance } from "../Config/Config";
-import { IGitLabWebhookMREvent, IGitLabWebhookPushEvent, IGitLabWebhookReleaseEvent, IGitLabWebhookTagPushEvent, IGitLabWebhookWikiPageEvent } from "../Gitlab/WebhookTypes";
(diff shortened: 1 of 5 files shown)


src/Connections/GitlabRepo.ts
```diff
@@ -292,6 +296,60 @@ ${data.description}`;
...
+        const mergeRequest = event.merge_request;
+
+        const renderFn = () => {
+            const result = this.debounceMRComments.get(uniqueId);
+            if (!result) {
+                // Always defined, but for type checking purposes.
+                return;
+            }
+            const orgRepoName = event.project.path_with_namespace;
+            const comments = result.comments > 1 ? `${result.comments} comments` : '1 comment';
```
jaller94, line 325: Can a review have 0 comments?
```suggestion
            const comments = result.comments !== 1 ? `${result.comments} comments` : '1 comment';
```

Half-Shot: Not really, the function can only be triggered by a comment...so you're going to have at least one.

jaller94 (approved):
Looks sensible to me.

Issue #180, opened 2022-01-20
Support RSS / ATOM feeds


Half-Shot:
Should be straight forward to do.

PR #315, opened 2022-04-21
Add RSS/Atom feed support

tadzik:
Fixes https://github.com/matrix-org/matrix-hookshot/issues/180


config.sample.yml
```diff
@@ -80,6 +80,10 @@ generic:
   userIdPrefix: _webhooks_
   allowJsTransformationFunctions: false
   waitForComplete: false
+feeds:
+  # (Optional) Support for RSS/Atom feeds.
+  enabled: false
+  pollIntervalSeconds: 600
```
HarHarLinks, line 86: it would be awesome if this was configurable on a per-feed basis, up (down?) to an admin configured minimum

src/Connections/FeedConnection.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    public isInterestedInStateEvent(eventType: string, stateKey: string): boolean {
+        return false;
+    }
+
+    public async handleFeedEntry(entry: FeedEntry): Promise<void> {
+        this.hasError = false;
+        const message = `New post in ${entry.feed.title}: [${entry.title}](${entry.link})`
+        await this.as.botIntent.sendEvent(this.roomId, {
+            msgtype: 'm.notice',
```
HarHarLinks, line 47: I would also be nice if you could receive notifications from feeds, if you select (i.e. configure this on a per-feed basis?)

HarHarLinks (commented):
👀

src/Connections/SetupConnection.ts
```diff
@@ -157,6 +135,25 @@ export class SetupConnection extends CommandConnection {
         return this.as.botClient.sendHtmlNotice(this.roomId, md.renderInline(`Room configured to bridge Figma file.`));
     }
 
+    @botCommand("feed", { help: "Bridge an RSS/Atom feed to the room.", requiredArgs: ["url"], includeUserId: true, category: "feed"})
+    public async onFeed(userId: string, url: string) {
+        if (!this.config.feeds?.enabled) {
+            throw new CommandError("not-configured", "The bridge is not configured to support feeds.");
+        }
+
+        await this.checkUserPermissions(userId, "figma", FeedConnection.CanonicalEventType);
```
Half-Shot, line 144:
```suggestion
        await this.checkUserPermissions(userId, "feed", FeedConnection.CanonicalEventType);
```

src/feeds/FeedReader.ts
```diff
@@ -0,0 +1,144 @@
...
+                    this.queue.push<FeedEntry>({ eventName: 'feed.entry', sender: 'FeedReader', data: entry });
+                }
+
+                if (seenEntriesChanged) {
+                    // Some RSS feeds can return a very small number of items then bounce
+                    // back to their "normal" size, so we cannot just clobber the recent GUID list per request or else we'll
+                    // forget what we sent and resend it. Instead, we'll keep 2x the max number of items that we've ever
+                    // seen from this feed, up to a max of 10,000. 
+                    // Adopted from https://github.com/matrix-org/go-neb/blob/babb74fa729882d7265ff507b09080e732d060ae/services/rssbot/rssbot.go#L304
+                    const maxGuids = Math.min(Math.max(2 * newGuids.length, seenGuids.length), 10_000);;
```
Half-Shot, line 129:
```suggestion
                    const maxGuids = Math.min(Math.max(2 * newGuids.length, seenGuids.length), 10_000);
```

src/Config/Config.ts
```diff
@@ -395,6 +403,7 @@ export class BridgeConfig {
         this.figma = configData.figma;
         this.jira = configData.jira && new BridgeConfigJira(configData.jira);
         this.generic = configData.generic && new BridgeConfigGenericWebhooks(configData.generic);
+        this.feeds = configData.feeds;
```
Half-Shot, line 406: Remember to update the check on line 440

src/Config/Config.ts
```diff
@@ -184,6 +184,11 @@ export class BridgeConfigGitLab {
     }
 }
 
+export interface BridgeConfigFeeds {
```
Half-Shot, line 187: Bonus points for updating the Config/Defaults.ts (which isn't really defaults, more of a sample config) so that users know how to configure it. Also make sure to regen the sample config.

src/ConnectionManager.ts
```diff
@@ -54,6 +56,7 @@ export class ConnectionManager {
         for (const connection of connections) {
             if (!this.connections.find(c => c.connectionId === connection.connectionId)) {
                 this.connections.push(connection);
+                this.emit('new-connection', connection);
```
Half-Shot, line 59: Possibly also want one for removed connections?

src/Connections/FeedConnection.ts
```diff
@@ -0,0 +1,68 @@
...
+    url: string;
+}
+
+export class FeedConnection extends BaseConnection implements IConnection {
+    static readonly CanonicalEventType = "uk.half-shot.matrix-hookshot.feed";
+    static readonly EventTypes = [ FeedConnection.CanonicalEventType ];
+    private hasError = false;
+
+    public get feedUrl(): string {
+        return this.state.url;
```
Half-Shot, line 24: I would put some linting code around here to trim out any crap from a user provided URL (e.g. `#` values), converting to lowercase, trimming etc. Hopefully this means we can dedupe URLs further on.

src/feeds/FeedReader.ts
```diff
@@ -0,0 +1,144 @@
...
+    title: string,
+    link:  string,
+}
+
+interface AccountData {
+    [url: string]: string[],
+}
+
+function stripHtml(input: string): string {
+    return input.replace(/<[^>]*?>/g, '');
```
Half-Shot, line 36: Heh, I'm surprised it's that easy. If it works it works though!

src/feeds/FeedReader.ts
```diff
@@ -0,0 +1,144 @@
...
+            return this.pollFeeds();
+        });
+    }
+
+    private async loadSeenEntries(): Promise<void> {
+        const accountData = await this.matrixClient.getAccountData<AccountData>(FeedReader.seenEntriesEventType).catch((err: any) => {
+            if (err.statusCode === 404) {
+                return {} as AccountData;
+            } else {
+                throw err;
```
Half-Shot, line 71: I'm not sure we're catching this. I suspect it might cause the application to crash as `loadSeenEntries` will throw and uncaught promises crash later versions of Node.

src/feeds/FeedReader.ts
```diff
@@ -0,0 +1,144 @@
...
+            }
+        });
+        for (const url in accountData) {
```
Half-Shot, line 74: Type-check here, since it's possible there have been some cheeky tamperings with the account data.

src/feeds/FeedReader.ts
```diff
@@ -0,0 +1,144 @@
...
+        }
+        await this.matrixClient.setAccountData(FeedReader.seenEntriesEventType, accountData);
+    }
+
+    private async pollFeeds(): Promise<void> {
+        log.debug(`Checking for updates in ${this.observedFeedUrls.length} RSS/Atom feeds`);
+
+        let seenEntriesChanged = false;
+
+        await Promise.all(this.observedFeedUrls.map(async (url) => {
```
Half-Shot, line 92: Hmm, this feels a bit too thundering herd-y to me. We could end up checking 1000s of feeds at once, and then do nothing for N seconds which would potentially be quite disruptive. Can we stagger them a bit?

Half-Shot (changes requested):
This is looking excellent, nice work :)

src/Connections/FeedConnection.ts
tadzik: I'm not sure what you mean by that. Do you mean to subscribe to a particular feed as an individual? That's something you should be able to do already by just creating an RSS room for yourself.

config.sample.yml
tadzik: I've pondered it but I feel like that has more edgecases than usecases :)

The `pollIntervalSeconds` is something that the server admin would set to strike a balance between the network traffic on their server (and how "nice" they are to feed hosts) and the responsiveness of the subscriptions.

If we treat it as a minimum interval, we're opening an option for individual subscribers to say "I want to be nicer than this" – which is valid if they don't care about responsiveness that much, but it only works if all subscribers to this feed decide to be nice – it only takes one eager subscriber to force us to refresh it at a minimum anyway.

If we treat it as a maximum on the other hand, any user setting it up can effectively break the contract that the server admin signed, so it's not something we'd want.

So we could treat it as a default and allow a range of intervals to use, but it seriously feels like it's more hassle than it's worth: configuration-, UX- and complexity-wise. Is there good usecase for this that I didn't think of? I'm happy to change my mind :)

src/Config/Config.ts
tadzik: Aha, good call. Fixed in 648e457

src/feeds/FeedReader.ts
tadzik: I was hoping there's already something in yarn.lock that I won't have to reimplement, but there isn't and I don't feel like pulling in another dependency for something this stupid :P I *think* it covers all the bases. We may want to strip markdown to, just in case, since we appear to interpret it later when posting the message to the room.

src/feeds/FeedReader.ts
tadzik: Ehh, right. I guess our only way out then is wipe it and start from scratch?

src/Connections/SetupConnection.ts
tadzik: Ugh. Fixed in 917ea00

src/feeds/FeedReader.ts
tadzik: Fixed in 917ea00.

src/Config/Config.ts
tadzik: Ah, I did that manually, didn't notice it's generated. Fixed in ef91798.

src/feeds/FeedReader.ts
tadzik: Adjusted in 420def0 – we now do it sequentially.

src/feeds/FeedReader.ts
Half-Shot: Yeah, pretty much.

src/Connections/FeedConnection.ts
HarHarLinks: I want to subscribe to a feed and select whether hookshot posts as `m.notice` (no notification) or `m.text` (notification). Then I can subscribe to "important" feeds and be explicitly notified when it updates.

src/Config/Config.ts
Half-Shot: We should get a contributing file that explains all the magic things we do.

config.sample.yml
HarHarLinks: You make a compelling argument, this might be unnecessary as a feature after all.

src/feeds/FeedReader.ts
tadzik: Fixed in 428ca024569010d0350663fcfa3a7b760ec17744. Brought in the entire `ajv` since we already depended on it :)

src/feeds/FeedReader.ts
tadzik: Changed it to be non-throwing in 428ca024569010d0350663fcfa3a7b760ec17744

src/ConnectionManager.ts
tadzik: Added now in a62235e925d810ce39059784ff7db7594517b22e and implemented in FeedConnection (which drove deduplicating URLs while we were at it).

src/Connections/FeedConnection.ts
tadzik: I've done it in FeedReader in a62235e925d810ce39059784ff7db7594517b22e, while `Set`ing them to ensure that we're not fetching a feed twice if it appears in two rooms. I stripped HTML anchors but left the casing intact, as `/feed.xml?secretToken=aaa` and /feed.xml?secretToken=AAA` are in fact two different URLs and potentially different feeds.

src/Connections/FeedConnection.ts
tadzik: Ah, I see – that's a good point.

I wonder if this should be a global hookshot feature, or at least implemented in a way that can be reused easily across different services. What do you think @Half-Shot?

tadzik: Built a prototype in https://github.com/matrix-org/matrix-hookshot/pull/316

Half-Shot: Happy for this to land across all connections in a different PR.

Half-Shot (commented):
Very close now, just a few things.

docs/setup/feeds.md
```diff
@@ -0,0 +1,17 @@
+# Feeds
```
Half-Shot, line 1: I added some skeleton docs because I couldn't resist. We have quite a few places that need updating when a new service is added so I thought this would save some time.

docs/metrics.md
```diff
@@ -30,6 +32,11 @@ Below is the generated list of Prometheus metrics for Hookshot.
 | matrix_api_calls | The number of Matrix client API calls made | method |
 | matrix_api_calls_failed | The number of Matrix client API calls which failed | method |
 | matrix_appservice_events | The number of events sent over the AS API |  |
+## feed
```
Half-Shot, line 35: Added a couple of useful metrics to watch the needle as we scale up.

src/feeds/FeedReader.ts
```diff
@@ -0,0 +1,226 @@
...
+                    }
+                    if (seenGuidsSet.has(guid)) {
+                        log.debug('Skipping already seen entry', guid);
+                        continue;
+                    }
+
+                    const entry = {
+                        feed: { title: stripHtml(feed.title!), url: url.toString() },
+                        title: stripHtml(item.title!),
+                        link: item.link!,
```
Half-Shot, line 183: Presumably any of these could be undefined right? Do we need to be careful here and check?

tadzik: Could happen, yeah. I made sure we're making sense of the edgecases in b8afd4f7ef5248054c1deb32286f0275612935f2

src/feeds/FeedReader.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    constructor(
+        private config: BridgeConfigFeeds,
+        private connectionManager: ConnectionManager,
+        private queue: MessageQueue,
+        private matrixClient: MatrixClient,
+    ) {
+        this.connections = this.connectionManager.getAllConnectionsOfType(FeedConnection);
+        this.calculateFeedUrls();
+        connectionManager.on('new-connection', c => {
```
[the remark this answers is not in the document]
tadzik: As discussed [on Matrix](https://matrix.to/#/!TlZdPIYrhwNvXlBiEk:half-shot.uk/$aYZYuUK0AGkUGeEqGcKYINmU_w67Xzotq9LQd_z2pXY?via=half-shot.uk&via=matrix.org&via=envs.net), we cannot trivially do it now and can consider if later if the need arises (probably by splitting FeedReader itself into two separate parts, one not aware of Connections as a thing and possibly running in a separate process).

docs/setup/feeds.md
tadzik: Fleshed them out now:)

Half-Shot (approved):
Seems reasonable, fix lint and then you can merge :)

PR #320, opened 2022-04-25
Add support for GitLab connection configuration via widgets

Half-Shot:
This PR has refactored a lot of what we did in the webhooks code to be more generic, hopefully allowing the next PR to be much much smaller. Either way, this is *feeling* like the right approach.

Specifically this PR enables support for creating and removing GitLab projects from rooms.

### Screenshots

Blank:

image

Adding a new connection:

image

New connection:

image

New connection with fewer events selected:

image


web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+    }, [filter, setFilter]);
+
+    useEffect(() => {
+        if (debounceTimer) {
+            clearTimeout(debounceTimer);
+        }
+        // Browser types
+        setDebounceTimer(setTimeout(searchFn, 500) as unknown as number);
+        // Things break if we depend on the thing we are clearing.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
jaller94, line 43: What kind of dark React magic is being performed here??
If it works, I'll pretend not to care, but this is how you'd usually clean up Effects:

You `return` a function that undoes your changes.
https://gitlab.com/jaller94/matrix-wrench/-/blob/9f045b8926802fde53c7097c8b26bb58d632b88d/src/index.js#L392-395

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }, [onPicked, results, filter, setCurrentProjectPath]);
+
+
+
+    return <div>
+        {instances === null && <p> Loading GitLab instances. </p>}
+        {instances?.length === 0 && <p> You are not logged into any GitLab instances. </p>}
+        <InputField visible={instances?.length > 0} label="GitLab Instance" noPadding={true}>
+            {/*TODO: on change required here*/}
```
jaller94, line 60: mhhh... sounds like a required todo left unsolved.
The select won't work, I guess.

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+            <select>
+                {instances?.map(i => <option key={i.name}>{i.name}</option>)}
+            </select>
+        </InputField>
+        <InputField visible={instances?.length > 0} label="Project" noPadding={true}>
+            <small>{currentProjectPath ?? ""}</small>
+            <input onChange={updateSearchFn} list="gitlab-projects" type="text" />
+            <datalist id="gitlab-projects">
+                {results?.map(i => <option path={i.state.path} value={i.name} key={i.name} />)}
```
jaller94, line 69: Even lists like this can be thrown into `useMemo` if they are expected to display long lists of 10s of items.

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }, [ignoredHooks, setIgnoredHooks]);
+    const [newInstanceState, setNewInstanceState] = useState<GitLabRepoConnectionState|null>(null);
+
+
+
+    const canEdit = !existingConnection || (existingConnection?.canEdit ?? false);
+    const commandPrefixRef = createRef<HTMLInputElement>();
+
+    const onSaveCb = useCallback(() => {
```
jaller94, line 92: Just a suggestion for naming these:
For several years I've adopted the following naming scheme.

`onSave` – the prop from a parent component
`handleSave` – the callback in the component (and likely calls `onSave`)

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+        api={api}
+        roomId={roomId}
+        type="gitlab"
+        text={({
+            header: 'GitLab Projects',
+            createNew: 'Add new GitLab project',
+            listCanEdit: 'Your connected projects',
+            listCantEdit: 'Connected projects',
+        })}
+        listItemName={(c) => (c as GitLabRepoResponseItem).config.path}
```
jaller94, line 163: Same as above.
```suggestion
        text={useMemo({
            header: 'GitLab Projects',
            createNew: 'Add new GitLab project',
            listCanEdit: 'Your connected projects',
            listCantEdit: 'Connected projects',
        }, [])}
        listItemName={useCallback((c) => (c as GitLabRepoResponseItem).config.path, [])}
```

web/components/roomConfig/RoomConfig.tsx
```diff
@@ -0,0 +1,124 @@
+import { h, FunctionComponent, Fragment } from "preact";
+import { useState } from "preact/hooks"
+import { ListItem } from "../ListItem";
+import BridgeAPI from "../../BridgeAPI";
+import ErrorPane from "../ErrorPane";
+import style from "./RoomConfig.module.scss";
+import { GetConnectionsResponseItem } from "../../../src/provisioning/api";
```
jaller94, line 7: Importing from `../../../src` looks like an import that should be illegal or you're ending up with the back-end code in your front-end bundle at some point.

web/components/roomConfig/RoomConfig.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    if (serviceConfig === null) {
+        api.getServiceConfig<SConfig>(type)
+        .then((res) => setServiceConfig(res))
+        .catch(ex => {
+            console.warn("Failed to fetch service config", ex);
+            setError("Failed to fetch service config");
+        });
+
+    }
```
jaller94, line 62: :warning: This needs fixing.
Creates an endless loop of network calls on failure.

1. Request faills and `setError` is called.
2. Component re-renders with `connections` and `serviceConfig` being `null`.
3. Request faills and `setError` is called.
4. …

The proper way to make network calls on component creation is:
```
useEffect(() => {
    // make your network calls here.
}, []); // The empty array means that this will only be executed once.
```

web/components/roomConfig/RoomConfig.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    // We need to increment this every time we create a connection in order to properly reset the state.
+    const [ newConnectionKey, setNewConnectionKey ] = useState<number>(0);
+
+    return <>
```
jaller94, line 67: No need for the outer most Fragment `<> </>`.
`<main></main>` is the only child.

web/components/roomConfig/RoomConfig.tsx
```diff
@@ -0,0 +1,124 @@
...
+                <img src={headerImg} />
+                <h1>{text.header}</h1> 
+            </header>
+            { canEditRoom && <section>
+                <h2>{text.createNew}</h2>
+                {serviceConfig && <ConnectionConfigComponent
+                    key={newConnectionKey}
+                    api={api}
+                    serviceConfig={serviceConfig}
+                    onSave={(config) => {
```
jaller94, line 82: This would benefit from `onCallback`, however it would need to be defined outside the conditional `{foo &&` clauses.

web/styling.scss
```diff
@@ -5,6 +5,7 @@
   --foreground-color: #17191C;
   --light-color: #737D8C;
   --primary-color: #0DBD8B;
+  --primary-color-disabled: #0dbd8bb4;
```
jaller94, line 8: TIL: Browsers accept 4 byte hex colours to also set the alpha value.

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+                }
+            } else {
+                setResults(res as GitLabRepoConnectionProjectTarget[]);
+            }
+        });
+    }, [api, filter]);
+
+    const updateSearchFn = useCallback((evt: InputEvent) => {
+        setFilter({...filter, search: (evt.target as HTMLInputElement).value });
+    }, [filter, setFilter]);
```
jaller94, line 34: Setters from `useState` never change.
```suggestion
    }, [filter]);
```

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [setDebounceTimer, searchFn, filter, instances]);
+
+    useEffect(() => {
+        const hasResult = results?.find(n => n.name === filter.search);
+        if (hasResult) {
+            onPicked(hasResult.state);
+            setCurrentProjectPath(hasResult.state.path);
+        }
+    }, [onPicked, results, filter, setCurrentProjectPath]);
```
jaller94, line 52: Setter from `useState` never change.
```suggestion
    }, [onPicked, results, filter]);
```

src/Widgets/BridgeWidgetApi.ts
```diff
@@ -180,4 +168,13 @@ export class BridgeWidgetApi {
         await this.connMan.purgeConnection(roomId, connectionId);
         res.send({ok: true});
     }
+
+    private async getConnectionTargets(req: ProvisioningRequest, res: Response) {
+        if (!req.userId) {
+            throw Error('Cannot get connections without a valid userId');
+        }
+        const type = req.params.type as string;
```
jaller94, line 176: If that's not guaranteed to be a `string` you shouldn't cast it to a string.

Do you expect `this.connMan.getConnectionTargets` to deal with this? Then please change its signature.

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+        type="gitlab"
+        text={({
+            header: 'GitLab Projects',
+            createNew: 'Add new GitLab project',
+            listCanEdit: 'Your connected projects',
+            listCantEdit: 'Connected projects',
+        })}
+        listItemName={(c) => (c as GitLabRepoResponseItem).config.path}
+        connectionEventType={EventType}
+        connetionConfigComponent={ConnectionConfiguration}
```
jaller94, line 165:
```suggestion
        connectionConfigComponent={ConnectionConfiguration}
```

web/components/roomConfig/RoomConfig.tsx
```diff
@@ -0,0 +1,124 @@
...
+    headerImg: string;
+    text: {
+        header: string;
+        createNew: string;
+        listCanEdit: string;
+        listCantEdit: string;
+    };
+    connectionEventType: string;
+    listItemName: (c: ConnectionType) => string,
+    connetionConfigComponent: FunctionComponent<ConnectionConfigurationProps<SConfig, ConnectionType, ConnectionState>>;
```
jaller94, line 31:
```suggestion
    connectionConfigComponent: FunctionComponent<ConnectionConfigurationProps<SConfig, ConnectionType, ConnectionState>>;
```

web/components/roomConfig/RoomConfig.tsx
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const RoomConfig: FunctionComponent<IRoomConfigProps<any, any, any>> = function<SConfig, ConnectionType extends GetConnectionsResponseItem, ConnectionState>(props: IRoomConfigProps<SConfig, ConnectionType, ConnectionState>) {
+    const { api, roomId, type, headerImg, text, listItemName, connectionEventType } = props;
+    const ConnectionConfigComponent = props.connetionConfigComponent;
```
jaller94, line 37:
```suggestion
    const ConnectionConfigComponent = props.connectionConfigComponent;
```

src/Connections/GitlabRepo.ts
```diff
@@ -512,6 +563,29 @@ ${data.description}`;
...
+        log.info(`Removing ${this.toString()} for ${this.roomId}`);
+        // Do a sanity check that the event exists.
+        try {
+            await this.as.botClient.getRoomStateEvent(this.roomId, GitLabRepoConnection.CanonicalEventType, this.stateKey);
+            await this.as.botClient.sendStateEvent(this.roomId, GitLabRepoConnection.CanonicalEventType, this.stateKey, { disabled: true });
+        } catch (ex) {
+            await this.as.botClient.getRoomStateEvent(this.roomId, GitLabRepoConnection.LegacyCanonicalEventType, this.stateKey);
+            await this.as.botClient.sendStateEvent(this.roomId, GitLabRepoConnection.LegacyCanonicalEventType, this.stateKey, { disabled: true });
+        }
+        // TODO: Clean up webhooks
```
jaller94, line 587: :eyes: What webhooks need to be cleaned up here?
Is this required for a merge?

src/Connections/GitlabRepo.ts
```diff
@@ -158,6 +177,48 @@ export class GitLabRepoConnection extends CommandConnection {
         }
     }
 
+    public static getProvisionerDetails(botUserId: string) {
+        return {
+            service: "gitlab",
+            eventType: GitLabRepoConnection.CanonicalEventType,
+            type: "GitLabRepo",
+            // TODO: Add ability to configure the bot per connnection type.
+            botUserId: botUserId,
```
jaller94, line 186:
```suggestion
            botUserId,
```

Also, unsolved TODO in the line above.

src/Connections/GitlabRepo.ts
```diff
@@ -158,6 +177,48 @@ export class GitLabRepoConnection extends CommandConnection {
...
+            }
+            return results;
+        }
+        // If we have an instance, search under it.
+        const instanceUrl = config.instances[filters.instance]?.url;
+        const client = instanceUrl && await tokenStore.getGitLabForUser(userId, instanceUrl);
+        if (!client) {
+            throw new ApiError('Instance is not known or you do not have access to it.', ErrCode.NotFound);
+        }
+        const after = filters.after === undefined ? undefined : parseInt(filters.after, 10);
```
jaller94, line 211: Aside from `??` also replacing `null` this seems to be what `??` was introduced for.
```suggestion
        const after = filters.after ?? parseInt(filters.after, 10);
```

src/Connections/GitlabRepo.ts
```diff
@@ -512,6 +563,29 @@ ${data.description}`;
         }
         return false;
     }
+
+    public async provisionerUpdateConfig(userId: string, config: Record<string, unknown>) {
+        const validatedConfig = validateState(config);
+        await this.as.botClient.sendStateEvent(this.roomId, GitLabRepoConnection.CanonicalEventType, this.stateKey, 
+            {
+                ...validatedConfig,
+            }
```
jaller94, line 572: `validatedConfig` already is a new object that's not handed to any other function.
It wouldn't need to be redefined here.

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+            setCurrentProjectPath(hasResult.state.path);
+        }
+    }, [onPicked, results, filter, setCurrentProjectPath]);
+
+
+
+    return <div>
+        {instances === null && <p> Loading GitLab instances. </p>}
+        {instances?.length === 0 && <p> You are not logged into any GitLab instances. </p>}
+        <InputField visible={instances?.length > 0} label="GitLab Instance" noPadding={true}>
```
jaller94, line 59: Why even render this with `visible = false` if there are no instances?

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+    return <div>
+        {!existingConnection && <ConnectionSearch api={api} onPicked={setNewInstanceState} />}
+        <InputField visible={!!existingConnection} label="GitLab Instance" noPadding={true}>
+            <input disabled={true} type="text" value={existingConnection?.config.instance} />
+        </InputField>
+        <InputField visible={!!existingConnection} label="Project" noPadding={true}>
+            <input disabled={true} type="text" value={existingConnection?.config.path} />
+        </InputField>
+        <InputField visible={!!existingConnection || !!newInstanceState} ref={commandPrefixRef} label="Command Prefix" noPadding={true}>
+            <input type="text" value={existingConnection?.config.commandPrefix || "!gl"} />
```
jaller94, line 109: Could empty string be a valid existing configuration for the Command Prefix?

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+                <ul>
+                    <li><input disabled={ignoredHooks.includes("merge_request")} type="checkbox" x-event-name="merge_request.open" checked={!ignoredHooks.includes("merge_request.open")} onChange={setIgnoredHook} />Opened</li>
+                    <li><input disabled={ignoredHooks.includes("merge_request")} type="checkbox" x-event-name="merge_request.close" checked={!ignoredHooks.includes("merge_request.close")} onChange={setIgnoredHook} />Closed</li>
+                    <li><input disabled={ignoredHooks.includes("merge_request")} type="checkbox" x-event-name="merge_request.merge" checked={!ignoredHooks.includes("merge_request.merge")} onChange={setIgnoredHook} />Merged</li>
+                    <li><input disabled={ignoredHooks.includes("merge_request")} type="checkbox" x-event-name="merge_request.review" checked={!ignoredHooks.includes("merge_request.review")} onChange={setIgnoredHook} />Reviews</li>
+                </ul>
+                <li>
+                    <input type="checkbox" x-event-name="push" checked={!ignoredHooks.includes("push")} onChange={setIgnoredHook} />
+                    Pushes
```
jaller94, line 123: If you wrap these in `<label/>`s, clicking on the text will toggle the checkbox.

You may want to see if the wrapping label messes with your styles.

```suggestion
                    <label>
                        <input type="checkbox" x-event-name="push" checked={!ignoredHooks.includes("push")} onChange={setIgnoredHook} />
                        Pushes
                    </label>
```

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+const ConnectionConfiguration: FunctionComponent<ConnectionConfigurationProps<unknown, GitLabRepoResponseItem, GitLabRepoConnectionState>> = ({api, existingConnection, onSave, onRemove }) => {
+    const [ignoredHooks, setIgnoredHooks] = useState<string[]>(existingConnection?.config.ignoreHooks || []);
+    const setIgnoredHook = useCallback(evt => {
+        const key = (evt.target as HTMLElement).getAttribute('x-event-name');
+        if (ignoredHooks.includes(key)) {
+            setIgnoredHooks(ignoredHooks.filter(k => k !== key));
+        } else {
+            setIgnoredHooks([...ignoredHooks, key]);
+        }
+    }, [ignoredHooks, setIgnoredHooks]);
```
jaller94, line 84:
```suggestion
    }, [ignoredHooks]);
```

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+            <datalist id="gitlab-projects">
+                {results?.map(i => <option path={i.state.path} value={i.name} key={i.name} />)}
+            </datalist>
+        </InputField>
+    </div>;
+}
+
+const ConnectionConfiguration: FunctionComponent<ConnectionConfigurationProps<unknown, GitLabRepoResponseItem, GitLabRepoConnectionState>> = ({api, existingConnection, onSave, onRemove }) => {
+    const [ignoredHooks, setIgnoredHooks] = useState<string[]>(existingConnection?.config.ignoreHooks || []);
+    const setIgnoredHook = useCallback(evt => {
```
jaller94, line 77: This function had me confused for a while. I expected it to be looking at `event.target.checked` but you use the component's state.

That's fine, but maybe rename the function.

```suggestion
    const toggleIgnoredHook = useCallback(evt => {
```

jaller94 (changes requested):

src/ConnectionManager.ts
```diff
@@ -125,6 +125,24 @@ export class ConnectionManager extends EventEmitter {
             this.push(res.connection);
             return res.connection;
         }
+        if (GitLabRepoConnection.EventTypes.includes(type)) {
+            if (!this.config.gitlab) {
+                throw Error('Generic hook support not supported');
```
jaller94, line 130: 1. foo **support** is not **supported**
2. This is GitLab repos, not Generic hooks.

```suggestion
                throw Error('GitLab support not configured');
```

btw. the copy for Github just says "Github not configured" as if Microsoft had forgotten to configure it.

src/Gitlab/Client.ts
```diff
@@ -1,9 +1,10 @@
 import axios from "axios";
 import { GitLabInstance } from "../Config/Config";
-import { GetIssueResponse, GetUserResponse, CreateIssueOpts, CreateIssueResponse, GetIssueOpts, EditIssueOpts, GetTodosResponse, EventsOpts, CreateIssueNoteOpts, CreateIssueNoteResponse, GetProjectResponse, ProjectHook, ProjectHookOpts } from "./Types";
+import { GetIssueResponse, GetUserResponse, CreateIssueOpts, CreateIssueResponse, GetIssueOpts, EditIssueOpts, GetTodosResponse, EventsOpts, CreateIssueNoteOpts, CreateIssueNoteResponse, GetProjectResponse, ProjectHook, ProjectHookOpts, AccessLevel, SimpleProject } from "./Types";
 import LogWrapper from "../LogWrapper";
 import { URLSearchParams } from "url";
 import UserAgent from "../UserAgent";
+import { toHash } from "ajv/dist/compile/util";
```
jaller94, line 7: 'toHash' is defined but never used

web/components/InputField.module.scss
```diff
@@ -46,10 +28,15 @@
...
-    align-items: flex-start;
-}
\ No newline at end of file
+    select {
+        border: 1px solid #E3E8F0;
+        box-sizing: border-box;
+        border-radius: 8px;
+        width: 100%;
+        font-size: 14px;
+        background: #FFF;
```
jaller94, line 38: You allow yourself to set colours outside of variables?

src/Gitlab/Client.ts
```diff
@@ -71,6 +72,29 @@ export class GitLabClient {
         }
     }
 
+    private async listProjects(minAccess: AccessLevel, inGroup?: ProjectId, idAfter?: number, search?: string): Promise<SimpleProject[]> {
+        try {
+            const path = inGroup ? `api/v4/groups/${getProjectId(inGroup)}/projects` : 'api/v4/projects';
```
jaller94, line 77: What is `inGroup`? The type `ProjectId` does not seem to make sense.

src/Gitlab/Client.ts
```diff
@@ -71,6 +72,29 @@ export class GitLabClient {
         }
     }
 
+    private async listProjects(minAccess: AccessLevel, inGroup?: ProjectId, idAfter?: number, search?: string): Promise<SimpleProject[]> {
```
jaller94, line 75: TODO: Support listing more than 50 projects (pagination), I guess?!

web/components/roomConfig/GenericWebhookConfig.tsx
```diff
@@ -83,13 +68,11 @@ const ConnectionConfiguration: FunctionComponent<{
                 }}
             />
             <p> See the <a target="_blank" rel="noopener noreferrer" href={DOCUMENTATION_LINK}>documentation</a> for help writing transformation functions </p>
-        </div>}
-
-        <div className={style.buttonSet}>
-            { canEdit && <Button onClick={onSaveClick}>{ existingConnection ? "Save" : "Add webhook"}</Button>}
+        </InputField>
+        <ButtonSet>
+            { canEdit && <Button onClick={onSaveClick}>{ existingConnection ? "Save" : "Add webhook" }</Button>}
```
jaller94, line 73: Because you're not using HTML's `<form>` (which would be awesome for keyboard accessibility, e.g. pressing Ctrl+Enter in an input field submits a form), all your buttons should have `type="button"`. The default is `type="submit"` and they will try to `submit` their parent `<form>`, if there is one.

Beause every input handling is done is JS, you could also add `type="button"` as a default to `Button.tsx`. Many React UI libs do this.

web/components/roomConfig/GenericWebhookConfig.tsx
```diff
@@ -45,36 +39,27 @@ const ConnectionConfiguration: FunctionComponent<{
...
-            <label>URL</label>
+        <InputField visible={!!existingConnection} label="URL" noPadding={true}>
             <input disabled={true} placeholder="URL hidden" type="text" value={existingConnection?.secrets?.url || ""} />
-        </div>}
+        </InputField>
 
-        { serviceConfig.allowJsTransformationFunctions && <div className={style.inputField}>
-            <label className={style.nopad}>Enable Transformation JavaScript</label>
+        <InputField visible={serviceConfig.allowJsTransformationFunctions} label="Enable Transformation JavaScript" noPadding={true}>
             <input disabled={!canEdit} type="checkbox" checked={transFnEnabled} onChange={() => setTransFnEnabled(!transFnEnabled)} />
```
jaller94, line 59: The first parameter for a `useState` set callback is the current value.
```suggestion
            <input disabled={!canEdit} type="checkbox" checked={transFnEnabled} onChange={useCallback(() => setTransFnEnabled(v => !v), [])} />
```

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+import BridgeAPI from "../../BridgeAPI";
+import { ConnectionConfigurationProps, RoomConfig } from "./RoomConfig";
+import { GitLabRepoConnectionState, GitLabRepoResponseItem, GitLabTargetFilter, GitLabRepoConnectionTarget, GitLabRepoConnectionProjectTarget, GitLabRepoConnectionInstanceTarget } from "../../../src/Connections/GitlabRepo";
+import InputField from "../InputField";
+import ButtonSet from "../ButtonSet";
+import { Button } from "../Button";
+
+const EventType = "uk.half-shot.matrix-hookshot.gitlab.repository";
+
+const ConnectionSearch: FunctionComponent<{api: BridgeAPI, onPicked: (state: GitLabRepoConnectionState) => void}> = ({api, onPicked}) => {
```
jaller94, line 12: The component has no error state and will say "Loading GitLab instances." if loading the connection targets via the api fails.

web/components/roomConfig/GitlabRepoConfig.tsx
jaller94, line 34:
```suggestion
    const updateSearchFn = useCallback((evt: InputEvent) => {
        setFilter(filter => ({...filter, search: (evt.target as HTMLInputElement).value }));
    }, []);
```

web/components/roomConfig/GitlabRepoConfig.tsx
jaller94, line 84:
```suggestion
    const toggleIgnoredHook = useCallback(evt => {
        const key = (evt.target as HTMLElement).getAttribute('x-event-name');
        setIgnoredHooks(ignoredHooks => (
            ignoredHooks.includes(key) ? ignoredHooks.filter(k => k !== key) : [...ignoredHooks, key]
        ));
    }, []);
```

web/components/roomConfig/RoomConfig.tsx
```diff
@@ -0,0 +1,124 @@
...
+                                    // Force reload
+                                    setConnections(null);
+                                }).catch(ex => {
+                                    console.warn("Failed to create connection", ex);
+                                    setError("Failed to create connection");
+                                });
+                            }}
+                            onRemove={() => {
+                                api.removeConnection(roomId, c.id).then(() => {
+                                    setConnections(connections.filter(conn => c.id !== conn.id));
```
jaller94, line 112: Async update using a potentially outdated `connections` :warning: 
```suggestion
                                    setConnections(connections => connections.filter(conn => c.id !== conn.id));
```

web/components/roomConfig/RoomConfig.tsx
jaller94: When changing this you'll also need to change `setConnection(null)` to reload the connections.

web/components/roomConfig/RoomConfig.tsx
```diff
@@ -0,0 +1,124 @@
...
+                <h2>{text.createNew}</h2>
+                {serviceConfig && <ConnectionConfigComponent
+                    key={newConnectionKey}
+                    api={api}
+                    serviceConfig={serviceConfig}
+                    onSave={(config) => {
+                        api.createConnection(roomId, connectionEventType, config).then(() => {
+                            // Force reload
+                            setConnections(null);
+                            setNewConnectionKey(newConnectionKey+1);
```
jaller94, line 86:
```suggestion
                            setNewConnectionKey(v => v+1);
```

src/Widgets/BridgeWidgetApi.ts
```diff
@@ -30,16 +30,20 @@ export class BridgeWidgetApi {
             expressApp,
             widgetTokenPrefix: "hookshot_",
             disallowedIpRanges: config.widgets?.disallowedIpRanges,
+            openIdOverride: {
+                "beefy": new URL("http://localhost:8008")
```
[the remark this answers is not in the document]
Half-Shot: This is me hacking around my local homeserver not supporting federated lookups. I will make this configurable.

web/components/roomConfig/GenericWebhookConfig.tsx
```diff
@@ -45,36 +39,27 @@ const ConnectionConfiguration: FunctionComponent<{
...
-            nameRef.current.value = "";
-            setTransFn(EXAMPLE_SCRIPT);
-            setTransFnEnabled(false);
-        }
-    }, [onSave, nameRef, transFn, setTransFnEnabled, setTransFn, existingConnection, transFnEnabled]);
+    }, [onSave, nameRef, transFn, existingConnection, transFnEnabled]);
 
     const onRemoveClick = useCallback(() => {
         onRemove();
     }, [onRemove]);
```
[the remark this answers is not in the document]
Half-Shot: Hm, I think there was logic here once upon a time which was executed upon onRemove, but it's not needed now. Removing.

web/components/roomConfig/GenericWebhookConfig.tsx
```diff
@@ -103,84 +86,19 @@ interface ServiceConfig {
...
+        api={api}
+        roomId={roomId}
+        type="generic"
+        connectionEventType="uk.half-shot.matrix-hookshot.generic.hook"
+        text={({
+            header: 'Generic Webhooks',
+            createNew: 'Create new webhook',
+            listCanEdit: 'Your webhooks',
+            listCantEdit: 'Configured webhooks',
+        })}
```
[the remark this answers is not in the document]
Half-Shot: It didn't like memo, but I can probably just use a const value right?

web/components/roomConfig/GenericWebhookConfig.tsx
```diff
@@ -103,84 +86,19 @@ interface ServiceConfig {
...
+        listItemName={(c) => (c as GenericHookResponseItem).config.name}
```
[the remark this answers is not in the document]
Half-Shot: I'm going to be cheeky again and use a const, that seems okay to do?

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -0,0 +1,167 @@
...
+            if (!filter.instance) {
+                setInstances(res);
+                if (res[0]) {
+                    setFilter({instance: res[0].name});
+                }
+            } else {
+                setResults(res as GitLabRepoConnectionProjectTarget[]);
+            }
+        });
+    }, [api, filter]);
```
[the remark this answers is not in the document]
Half-Shot: Good shout, fixed.

web/components/roomConfig/GitlabRepoConfig.tsx
Half-Shot: Good shout on the return, have done that.

On the dark magic, I'm basically trying to write a debounce fn that clears the existing timeout when this effect is fired. The problem with including the `debounceTimer` value is that when we set the value of `debounceTimer`, the effect is presumably recalled leading to a loop?

If I am wrong on that one, we can probably drop this ignore line.

web/components/roomConfig/GitlabRepoConfig.tsx
Half-Shot: Now new and implemented!

web/components/roomConfig/GitlabRepoConfig.tsx
Half-Shot: Good shout, included. I wonder how many of these could be fixed this way.

web/components/roomConfig/GitlabRepoConfig.tsx
Half-Shot: I like that, will use it.

web/components/roomConfig/RoomConfig.tsx
Half-Shot: I *believe* this is fine as long as we're importing interfaces and not classes, at compile time this shouldn't get bundled.

Now we should probably have a separate package for interfaces to avoid this ever being contaminated though.

web/styling.scss
Half-Shot: This seemed to be what vs-code suggested to me, but it might be super incompatible.

src/Widgets/BridgeWidgetApi.ts
Half-Shot: I have no idea why I decided to cast this...it is a string.

src/Connections/GitlabRepo.ts
Half-Shot: We set up GitLab webhooks when setting up the connection, but we also need to remove them when they are no longer in use.

The trouble here is that we don't have context on whether other rooms are also using the same webhooks, so we don't want to vape them in case they are still used. I'll write some code for this.

src/Connections/GitlabRepo.ts
Half-Shot: I'll remove the TODO, it's basically a feature request and I think having these lines makes it more confusing.

src/Connections/GitlabRepo.ts
Half-Shot: I don't think so? `filters.after ?? foo` would return `foo` if the value *is* undefined/null rather than the opposite as we're doing here. I think I meant to use &&.

src/Connections/GitlabRepo.ts
Half-Shot: Whoops, c&p fail.

web/components/roomConfig/GitlabRepoConfig.tsx
Half-Shot: Not sure I follow? This doesn't render if there are no instances. The visible thing is a shorthand to avoid doing `{ doWeWantToShowThis && <input` everywhere, since it's a common need for input fields.

web/components/roomConfig/GitlabRepoConfig.tsx
Half-Shot: Empty string is indeed valid, I'll relegate "!gl" as a placeholder.

web/components/roomConfig/GitlabRepoConfig.tsx
Half-Shot: Done this, and also turned these into functional components to make this more readable.

src/ConnectionManager.ts
Half-Shot: Have tried to improve the wording for all.

web/components/InputField.module.scss
Half-Shot: This was definitely a bit lazy..fixing.

src/Gitlab/Client.ts
```diff
@@ -71,6 +72,29 @@ export class GitLabClient {
...
+                    pagination: "keyset",
+                    per_page: 50,
+                    order_by: "id",
+                    sort: "asc",
+                    id_after: idAfter,
+                    search,
+                }
+                })).data;
+        } catch (ex) {
+            log.warn(`Failed to get project:`, ex);
```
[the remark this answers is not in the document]
Half-Shot: I think this is okay, the log line already contains the context of the module it's being used in?

src/Gitlab/Client.ts
Half-Shot: `ProjectId`  was a messy way to support path segments (string[]), paths (string) and ids (number). It's a total mess though so I've cleaned it up.

This specific line is changing the API call depending on whether we are searching within a group for some projects, or searching globally.

web/components/roomConfig/GenericWebhookConfig.tsx
Half-Shot: I've made this a form instead, and hopefully done it right.

web/components/roomConfig/RoomConfig.tsx
Half-Shot: I used a reducer...which I think is right?

src/Config/Config.ts
```diff
@@ -257,6 +266,7 @@ interface BridgeWidgetConfigYAML {
     branding?: {
         widgetTitle: string,
     }
+    openIdOverrides?: Record<string, string>;
```
jaller94, line 269: I think the proper value type for most `Records` includes `undefined`.
TypeScript won't add this automatically, but obviously most Objects will return you `undefined` on many keys.

```suggestion
    openIdOverrides?: Record<string, string|undefined>;
```

src/Config/Config.ts
```diff
@@ -269,6 +279,8 @@ export class BridgeWidgetConfig {
     public readonly branding: {
         widgetTitle: string,
     }
+    @configKey("For testing only: A set of homeserver servernames mapped to their Server-Server API endpoints. This can be used if your local homeserver doesn't lookup via SRV or .well-known.")
+    public readonly openIdOverrides?: Record<string, URL>;
```
jaller94, line 283: Same here. The value could be a `URL`, but it could also be `undefined`.

src/Connections/GitlabRepo.ts
```diff
@@ -109,8 +128,9 @@ export class GitLabRepoConnection extends CommandConnection {
 
     private readonly debounceMRComments = new Map<string, {comments: number, author: string, timeout: NodeJS.Timeout}>();
 
-    public static async provisionConnection(roomId: string, requester: string, data: Record<string, unknown>, as: Appservice, tokenStore: UserTokenStore, instance: GitLabInstance, gitlabConfig: BridgeConfigGitLab) {
+    public static async provisionConnection(roomId: string, requester: string, data: Record<string, unknown>, as: Appservice, tokenStore: UserTokenStore, instanceName: string, gitlabConfig: BridgeConfigGitLab) {
         const validData = validateState(data);
+        const instance = gitlabConfig.instances[instanceName];
         const client = await tokenStore.getGitLabForUser(requester, instance.url);
```
jaller94, line 134: This is another object that's likely not guaranteeing to return an instance object, but could also return `undefined`.
The later would break the lookup of `.url` and throw an error.

web/components/roomConfig/GenericWebhookConfig.tsx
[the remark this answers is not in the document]
jaller94: Yes, but the const value need to be defined outside the function (to not be redefined on every call).

web/components/roomConfig/GitlabRepoConfig.tsx
jaller94: Ah, I see.. You've intentionally added a `visible` prop to some components to make the syntax of the parent component cleaner.

I like `{foo &&` better, but I see that it's mainly personal preference of what's considered cleaner.

web/components/roomConfig/RoomConfig.tsx
```diff
@@ -0,0 +1,127 @@
...
+    useEffect(() => {
+        const fn = (async () => {
+        try {
+            const res = await api.getServiceConfig<SConfig>(type);
+            setServiceConfig(res);
+        } catch (ex) {
+            console.warn("Failed to fetch service config", ex);
+            setError("Failed to fetch service config");
+        }
+        });
```
jaller94, line 67:
```suggestion
        const fn = async () => {
            try {
                const res = await api.getServiceConfig<SConfig>(type);
                setServiceConfig(res);
            } catch (ex) {
                console.warn("Failed to fetch service config", ex);
                setError("Failed to fetch service config");
            }
        };
```

web/components/roomConfig/GenericWebhookConfig.tsx
```diff
@@ -83,14 +68,12 @@ const ConnectionConfiguration: FunctionComponent<{
...
-
-        <div className={style.buttonSet}>
-            { canEdit && <Button onClick={onSaveClick}>{ existingConnection ? "Save" : "Add webhook"}</Button>}
-            { canEdit && existingConnection && <Button intent="remove" onClick={onRemoveClick}>Remove webhook</Button>}
-        </div>
-
-    </div>;
+        </InputField>
+        <ButtonSet>
+            { canEdit && <Button>{ existingConnection ? "Save" : "Add webhook" }</Button>}
```
jaller94, line 73: (Are you testing your changes? :eyes:)

For this to submit the form it must become a `submit` button.

```suggestion
            { canEdit && <Button type="submit">{ existingConnection ? "Save" : "Add webhook" }</Button>}
```

config.sample.yml
```diff
@@ -146,6 +146,10 @@ widgets:
   publicUrl: http://example.com/widgetapi/v1/static
   branding:
     widgetTitle: Hookshot Configuration
+  openIdOverrides:
+    # For testing only: A set of homeserver servernames mapped to their Server-Server API endpoints. This can be used if your local homeserver doesn't lookup via SRV or .well-known.
+    #
+    my-local-server: {}
```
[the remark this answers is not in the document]
Half-Shot: This has been dropped from this PR and now lives in #326

docs/advanced/widgets.md
```diff
@@ -42,6 +42,8 @@ widgets:
   publicUrl: http://example.com/widgetapi/v1/static
   branding:
     widgetTitle: Hookshot Configuration
+#  openIdOverrides:
+#    my-local-server: {}
```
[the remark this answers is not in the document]
Half-Shot: This has been dropped from this PR and now lives in #326

src/ConnectionManager.ts
```diff
@@ -419,6 +438,8 @@ export class ConnectionManager extends EventEmitter {
         if (requireNoRemoveHandler && !connection.onRemove) {
             throw Error("Connection doesn't support removal, and so cannot be safely removed");
         }
+        const prototype = Object.getPrototypeOf(connection);
```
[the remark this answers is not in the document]
Half-Shot: Yeah, oops. will remove.

web/components/roomConfig/GitlabRepoConfig.tsx
Half-Shot: Yeah, I suppose it is more uniform. Maybe I'm a bit of a holdout from the Angular era :), either way, let's leave it as it is for the moment.

web/components/roomConfig/GenericWebhookConfig.tsx
Half-Shot: Fixed, and yep now tested.

src/Connections/GitlabRepo.ts
```diff
@@ -512,6 +566,26 @@ ${data.description}`;
         }
         return false;
     }
+
+    public async provisionerUpdateConfig(userId: string, config: Record<string, unknown>) {
+        const validatedConfig = validateState(config);
+        await this.as.botClient.sendStateEvent(this.roomId, GitLabRepoConnection.CanonicalEventType, this.stateKey, validatedConfig
+        );
```
jaller94, line 573: Was the linter's max-len limit reached for the last two characters?
```suggestion
        await this.as.botClient.sendStateEvent(this.roomId, GitLabRepoConnection.CanonicalEventType, this.stateKey, validatedConfig);
```

src/Connections/IConnection.ts
```diff
@@ -63,6 +63,8 @@ export interface IConnection {
     /**
      * If supported, this is sent when a user attempts to remove the connection from a room. The connection
      *  state should be removed and any resources should be cleaned away.
+     * @props purgeRemoteConfig Should the remote configuration for the connection be purged (in the case that
+     * other connections may be sharing a remote resource).
```
jaller94, line 67: What does this have to do with anything?
Is that a `param` that's not in the function's signature?

src/Gitlab/Client.ts
```diff
@@ -58,22 +57,45 @@ export class GitLabClient {
...
+                    min_access_level: minAccess,
+                    simple: true,
+                    pagination: "keyset",
+                    per_page: 50,
+                    order_by: "id",
+                    sort: "asc",
+                    id_after: idAfter,
+                    search,
+                }
+                })).data;
```
jaller94, line 89: Is the linter inactive or `intent` not configured properly?
```suggestion
            })).data;
```

jaller94 (approved):
I think we've reviewed this enough.
Given that you're saying it works in testing, I think this is good to be merged.

PR #326, opened 2022-04-29
Allow overriding the results of an openID lookup for widgets for local development.

Half-Shot:
This is a config option I'd only expect to be used in dev environments, but it basically allows you to map a server hostname to a federation S-S API path for local dev so you don't need a working federated environment to test changes. It's a bit niche, but I've found it useful.

### DIFF
--- a/config.sample.yml
+++ b/config.sample.yml
@@ -146,6 +146,10 @@ widgets:
   publicUrl: http://example.com/widgetapi/v1/static
   branding:
     widgetTitle: Hookshot Configuration
+  openIdOverrides:
+    # For testing only: A set of homeserver servernames mapped to their Server-Server API endpoints. This can be used if your local homeserver doesn't lookup via SRV or .well-known.
+    #
+    my-local-server: {}
 permissions:
   # (Optional) Permissions for using the bridge. See docs/setup.md#permissions for help
   #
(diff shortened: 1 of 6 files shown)


config.sample.yml
```diff
@@ -146,6 +146,10 @@ widgets:
   publicUrl: http://example.com/widgetapi/v1/static
   branding:
     widgetTitle: Hookshot Configuration
+  openIdOverrides:
+    # For testing only: A set of homeserver servernames mapped to their Server-Server API endpoints. This can be used if your local homeserver doesn't lookup via SRV or .well-known.
+    #
+    my-local-server: {}
```
tadzik, line 152: The empty object here doesn't match what Config/Defaults seems to set. What's the format here, and what should one put here for it to work? Does `my-local-server` specifically have any meaning?

tadzik (changes requested):

config.sample.yml
tadzik: I still don't know what the structure is supposed to be :/ Locally I did a 

```
openIdOverrides:
    home.tadzik.net: "http://home.tadzik.net:8080"
```

And that seems to work, but that's not what the documented structure is – it is a `string` => `object` mapping in the docs, while `string` => `string` is what the code expects (and what seems to work).

tadzik (changes requested):
Still need clarifications

tadzik (approved):
Now it all fits :)

apoorv569:
@Half-Shot What about production servers that have disabled federation completely? Do we put `DOMAIN: https://DOMAIN`? that still wouldn't work I suppose. So what should we put here?

PR #331, opened 2022-05-04
Fix a typo in the README

anoadragon453:
I've tested this change for the past several answers and I *think* it's OK, but please double-check!!


anoadragon453:
Looks like I didn't test hard enough :(((

(assuming a changelog isn't necessary for this one? :innocent:)

tadzik (approved):
No unit tests, but I think this is safe.

PR #332, opened 2022-05-05
Move components from root to elements

Half-Shot:
We now have a sizeable amount of smaller React components which should really go in their own directory. I couldn't think of a better name than "elements" though I'm sure there probably is one.


tadzik (approved):
I can't think of a better name either, and a little tidyness won't hurt us :)

PR #346, opened 2022-05-09
Fix GitHub/GitLab state config not being honoured

### LINKED ISSUE #344: GitHub includingLabels option seems to no longer be filtering

Half-Shot:
Fixes #344 

This was broken in https://github.com/matrix-org/matrix-hookshot/pull/330 when we started using the output of the validation function as the config passed to the connections, so any configuration not handled by that validation function would be forgotten about. This PR:

- Uses ajv to ensure that we validate every expected property
- Adds some tests to cover a small amount of GitHubRepoConnection, enough to know that this bug is fixed.


src/Connections/GithubRepo.ts
```diff
@@ -107,6 +87,119 @@ const AllowedEvents: AllowedEventsNames[] = [
     "release",
 ];
 
+const ConnectionStateSchema = {
+  type: "object",
+  properties: {
+    priority: {
+        type: "number",
+        nullable: true,
```
Half-Shot, line 95: This would only compile after peppering the schema will these `nullables`. I suppose it's correct, but it's also naff.

src/Connections/GithubRepo.ts
```diff
@@ -115,46 +204,29 @@ function compareEmojiStrings(e0: string, e1: string, e0Index = 0) {
...
+	static validateState(state: Record<string, unknown>, isExistingState = false): GitHubRepoConnectionState {
+        const validator = new Ajv().compile(ConnectionStateSchema);
+        if (validator(state)) {
+            // Validate ignoreHooks IF this is an incoming update (we can be less strict for existing state)
+            if (!isExistingState && state.ignoreHooks && !state.ignoreHooks.every(h => AllowedEvents.includes(h))) {
+                throw new ApiError('`ignoreHooks` must only contain allowed values', ErrCode.BadValue);
+            }
+            return state;
+        }
+        throw new ValidatorApiError(validator.errors || []);
```
tadzik, line 222: Curious – is there an expected scenario when validation wouldn't pass but `validator.errors` would be empty? The code is sound regardless, I just wonder if it'd happen for anything other than ajv bug.

src/Connections/GithubRepo.ts
```diff
@@ -115,46 +204,29 @@ function compareEmojiStrings(e0: string, e1: string, e0Index = 0) {
...
 /**
  * Handles rooms connected to a github repo.
  */
 @Connection
 export class GitHubRepoConnection extends CommandConnection implements IConnection {
+
+	static validateState(state: Record<string, unknown>, isExistingState = false): GitHubRepoConnectionState {
+        const validator = new Ajv().compile(ConnectionStateSchema);
+        if (validator(state)) {
+            // Validate ignoreHooks IF this is an incoming update (we can be less strict for existing state)
```
tadzik, line 216: Are we only less strict then because of backwards compatibility?

src/Connections/GitlabRepo.ts
```diff
@@ -128,11 +144,23 @@ export class GitLabRepoConnection extends CommandConnection {
     static helpMessage: (cmdPrefix?: string | undefined) => MatrixMessageContent;
     static ServiceCategory = "gitlab";
 
+	static validateState(state: Record<string, unknown>, isExistingState = false): GitLabRepoConnectionState {
+        const validator = new Ajv({ strict: false }).compile(ConnectionStateSchema);
+        if (validator(state)) {
+            // Validate ignoreHooks IF this is an incoming update (we can be less strict for existing state)
```
tadzik, line 150: In this file this is already enforced by the JSON schema. Is this bit copypasted, or is the schema unnecessarily strict?

tests/utils/UserTokenStoreMock.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Api } from "@octokit/plugin-rest-endpoint-methods/dist-types/types";
+import { GitLabClient } from "../../src/Gitlab/Client";
+import { JiraClient } from "../../src/Jira/Client";
+import { JiraOAuth } from "../../src/Jira/OAuth";
+import { UserTokenStore } from "../../src/UserTokenStore";
+
+export class UserTokenStoreMock {
+    public jiraOAuth?: JiraOAuth;
+
+    public load(): Promise<void> {
```
tadzik, line 13: Any reason why all of these are defined at all (since they throw anyway)?

Seems like this entire Mock is actually unused (see the GithubRepoTest suggestion).

tests/connections/GithubRepoTest.ts
```diff
@@ -0,0 +1,147 @@
...
+	const githubInstance = new GithubInstance("foo", "bar");
+	const connection = new GitHubRepoConnection(
+		ROOM_ID,
+		as,
+		GitHubRepoConnection.validateState({
+			org: "a-fake-org",
+			repo: "a-fake-repo",
+			...state,
+		}),
+		tokenStore,
```
tadzik, line 56:
```suggestion
		{} as UserTokenStore,
```
This seems to work as well, and we don't need to create (and define) a mock at all :)

src/Connections/GithubRepo.ts
tadzik: This isn't covered by the new tests (regardless of `isExistingState`) – would be nice to have it there, and it'd be a good place for the rationale for the future reader :)

[instanbul](https://istanbul.js.org/) confirms:

image

tadzik (changes requested):
Fantastic – would be great to have the tests for Gitlab as well, but I'm not going to too picky since this is already a massive improvement. Left a few nits here and there.

src/Connections/GithubRepo.ts
tadzik: (And if that's valid ajv behaviour, it may be worth teaching ValidatorApiError's constructor to accept falsy `validator.errors` too, so that a future developer doesn't need to remember about this possibility.

src/Connections/GithubRepo.ts
Half-Shot: >Are we only less strict then because of backwards compatibility?

Yeah, though I'm wondering if it's even worth being strict here.

src/Connections/GithubRepo.ts
Half-Shot: I think this was just a typing bug yeah. 

>(And if that's valid ajv behaviour, it may be worth teaching ValidatorApiError's constructor to accept falsy validator.errors too, so that a future developer doesn't need to remember about this possibility.

Yep, let's do that.

src/Connections/GitlabRepo.ts
Half-Shot: Think I just failed to commit this actually.

tadzik (approved):

PR #364, opened 2022-05-24
Support GitHub Enterprise

### LINKED ISSUE #112: Github Enterprise support
Currently, as far as I can see only github.com is supported. However, Github Enterprise also has API & Github Apps (https://docs.github.com/en/enterprise-server@3.3/rest/guides/getting-started-with-the-rest-api), so at least on a quick glance, support for Github Enterprise might be possible technically.

Half-Shot:
Fixes #112


config.sample.yml
```diff
@@ -36,6 +40,7 @@ github:
     # (Optional) Prefix used when creating ghost users for GitHub accounts.
     #
     _github_
+  baseUrl: {}
```
tadzik, line 43: I don't think this one should be here.

tadzik (approved):
Looks good from here, save from an extra sample config key.

Issue #365, opened 2022-05-28
Error when connecting to GitHub


hwittenborn:
**Describe the bug**
The bot keeps exiting with a 404 when connecting to `https://github.com/api/v3/app`.

**Server Logs**
```jsonnet
INFO 20:24:48:362 [Bridge] Initialising memory storage
INFO 20:24:48:366 [Bridge] Starting up
INFO 20:24:48:366 [Bridge] Connecting to homeserver and fetching joined rooms..
INFO 20:24:48:414 [Config] Prefilling room membership for permissions for 0 rooms
ERROR 20:24:48:469 [App] BridgeApp encountered an error and has stopped: RequestError [HttpError]: Not Found
    at /bin/matrix-hookshot/node_modules/@octokit/request/dist-node/index.js:86:21
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async hook (/bin/matrix-hookshot/node_modules/@octokit/auth-app/dist-node/index.js:409:18)
    at async GithubInstance.start (/bin/matrix-hookshot/Github/GithubInstance.js:93:28)
    at async Bridge.start (/bin/matrix-hookshot/Bridge.js:86:13)
    at async start (/bin/matrix-hookshot/App/BridgeApp.js:38:5) {
  status: 404,
  response: {
    url: 'https://github.com/api/v3/app',
    status: 404,
    headers: {
      'cache-control': 'no-cache',
      connection: 'close',
      'content-encoding': 'gzip',
      'content-security-policy': "default-src 'none'; base-uri 'self'; connect-src 'self'; form-action 'self'; img-src 'self' data:; script-src 'self'; style-src 'unsafe-inline'",
      'content-type': 'text/plain; charset=utf-8',
      date: 'Sat, 28 May 2022 20:24:48 GMT',
      'expect-ct': 'max-age=2592000, report-uri="https://api.github.com/_private/browser/errors"',
      'permissions-policy': 'interest-cohort=()',
      'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',
      server: 'GitHub.com',
      'set-cookie': 'logged_in=no; domain=.github.com; path=/; expires=Sun, 28 May 2023 20:24:48 GMT; secure; HttpOnly; SameSite=Lax',
      'strict-transport-security': 'max-age=31536000; includeSubdomains; preload',
      'transfer-encoding': 'chunked',
[...]

PR #366, opened 2022-05-30
Fix Github API URLs

tadzik:
Fixes GH-365

### DIFF
--- a/src/Config/Config.ts
+++ b/src/Config/Config.ts
@@ -86,7 +86,7 @@ export class BridgeConfigGitHub {
         this.oauth = yaml.oauth;
         this.defaultOptions = yaml.defaultOptions;
         this.userIdPrefix = yaml.userIdPrefix || "_github_";
-        this.baseUrl = new URL(yaml.enterpriseUrl ?? "https://github.com");
+        this.baseUrl = new URL(yaml.enterpriseUrl ?? "https://api.github.com");
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/Github/GithubInstance.ts
```diff
@@ -42,7 +42,7 @@ export class GithubInstance {
     public static baseOctokitConfig(baseUrl: URL) {
         return {
             userAgent: UserAgent,
-            baseUrl: baseUrl && new URL("/api/v3", baseUrl).toString(),
+            baseUrl: baseUrl.toString().substring(0, baseUrl.toString().length-1),
```
jaller94, line 45: What is this doing? Removing the last character without checking what it is?

jaller94 (changes requested):
I don't understand the change and think it's incomplete.

src/Github/GithubInstance.ts
Half-Shot: Have added comments to explain how we do things, and removed the horrible URL mangling which wasn't required.

src/Config/Config.ts
```diff
@@ -86,7 +86,7 @@ export class BridgeConfigGitHub {
         this.oauth = yaml.oauth;
         this.defaultOptions = yaml.defaultOptions;
         this.userIdPrefix = yaml.userIdPrefix || "_github_";
-        this.baseUrl = new URL(yaml.enterpriseUrl ?? "https://github.com");
+        this.baseUrl = new URL(yaml.enterpriseUrl ?? "https://api.github.com");
```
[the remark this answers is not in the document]
Half-Shot: Fixed, hopefully made clearer.

PR #367, opened 2022-06-06
Add Admin command from listing and disconnecting connections

tadzik:
(Only lists Github for now).

image

### DIFF
--- a/src/AdminRoom.ts
+++ b/src/AdminRoom.ts
@@ -156,8 +156,12 @@ export class AdminRoom extends AdminRoomCommandHandler {
 
         // it's stupid that we need roomId -- shouldn't `id` identify the connection?
         const conn = this.connectionManager.getConnectionById(roomId, id);
+        if (!conn) {
+            await this.sendNotice("Connection not found");
+            return;
+        }
         try {
-            await this.connectionManager.purgeConnection(roomId, id);
+            await this.connectionManager.purgeConnection(conn.roomId, conn.connectionId);
             await this.sendNotice('Connection removed successfully');
         } catch (err: unknown) {
             log.debug(`Failed to purge connection: ${err}`);
@@ -230,8 +234,8 @@ export class AdminRoom extends AdminRoomCommandHandler {
         };
 
         const reposFormatted = connections.repos.map(c => ` - ${c.org}/${c.repo} (ID: \`${c.connectionId}\`, Room: \`${c.roomId}\`)`).join('\n');
-        const issuesFormatted = connections.issues.map(c => ` - ${c.org}/${c.repo}/${c.issueNumber} (ID: \`${c.connectionId}\`), Room: \`${c.roomId}\``).join('\n');
-        const discussionsFormatted = connections.discussions.map(c => ` - ${c.owner}/${c.repo} (ID: \`${c.connectionId}\`), Room: \`${c.roomId}\``).join('\n');
+        const issuesFormatted = connections.issues.map(c => ` - ${c.org}/${c.repo}/${c.issueNumber} (Room: \`${c.roomId}\`, ID: \`${c.connectionId}\`)`).join('\n');
+        const discussionsFormatted = connections.discussions.map(c => ` - ${c.owner}/${c.repo} (Room: \`${c.roomId}\`, ID: \`${c.connectionId}\`)`).join('\n');
 
         const content = [
             connections.repos.length > 0       ? `Repositories:\n${reposFormatted}`      : '',
--- a/src/AdminRoom.ts
+++ b/src/AdminRoom.ts
@@ -139,7 +139,7 @@ export class AdminRoom extends AdminRoomCommandHandler {
     @botCommand("help", { help: "This help text" })
     public async helpCommand() {
         const enabledCategories = [
-            this.canManageConnections() ? Category.ConnectionManagement : '',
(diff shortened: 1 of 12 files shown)


src/AdminRoom.ts
```diff
@@ -126,16 +130,30 @@ export class AdminRoom extends AdminRoomCommandHandler {
...
             this.config.github ? "github" : "",
             this.config.gitlab ? "gitlab" : "",
             this.config.jira ? "jira" : "",
         ];
         return this.botIntent.sendEvent(this.roomId, AdminRoom.helpMessage(undefined, enabledCategories));
     }
 
+    @botCommand("disconnect", { help: "Remove a connection", requiredArgs: ['roomId', 'id'], category: "general" })
+    public async disconnect(roomId: string, id: string) {
+        // it's stupid that we need roomId -- shouldn't `id` identify the connection?
```
tadzik, line 146: ID is derived from `roomId` *and* it should be unique, so aside of theoretical md5 conflicts we should just be able to change this API, I think.

src/Bridge.ts
```diff
@@ -1142,7 +1142,7 @@ export class Bridge {
 
     private async setUpAdminRoom(roomId: string, accountData: AdminAccountData, notifContent: NotificationFilterStateContent) {
         const adminRoom = new AdminRoom(
-            roomId, accountData, notifContent, this.as.botIntent, this.tokenStore, this.config,
+            roomId, accountData, notifContent, this.as.botIntent, this.tokenStore, this.config, this.connectionManager!,
```
Half-Shot, line 1145: Never use non-null here, always prefer a
```
if (!this.connectionManager) {
    throw Error('Expected connectionManager to be ready');
}
```

src/AdminRoom.ts
Half-Shot: I'd be up for that change, I'd imagine conflicts are fairly rare here. We should explode in fire if there *IS* a conflict though so we don't end up deleting the wrong one (and they can file a bug report with us, or delete it manually).

src/AdminRoom.ts
```diff
@@ -126,16 +130,30 @@ export class AdminRoom extends AdminRoomCommandHandler {
...
         const enabledCategories = [
+            "general",
             this.config.github ? "github" : "",
             this.config.gitlab ? "gitlab" : "",
             this.config.jira ? "jira" : "",
         ];
         return this.botIntent.sendEvent(this.roomId, AdminRoom.helpMessage(undefined, enabledCategories));
     }
 
+    @botCommand("disconnect", { help: "Remove a connection", requiredArgs: ['roomId', 'id'], category: "general" })
```
Half-Shot, line 144: How are we securing these endpoints against not-admins? The admin room is slightly badly named in that it's your personal administration room, not always the bridge admin. We have a bridge admin role in the config that we could use to check with though.

Half-Shot (changes requested):
Needs a bit more thought but on the right track

src/AdminRoom.ts
tadzik: > The admin room is slightly badly named in that it's your personal administration room

Oops. Seems like I got bitten by this :)

Addressed in 8842b893751da8eff5a38a045f71605523887bab which requires `connectionManager` permission level (which conveniently exists), also giving these their own config section.

src/Bridge.ts
tadzik: Changed in b27e9657694b5ee09aa6b3537d76783a79745ee0.

One thing I often use instead in situations like this (with a lack of rxjs' `AsyncSubject`), is having this stored as a `Promise<ConnectionManager>`, initialized in the constructor and `await`ed everywhere it's needed – at the cost of some async-complexity this makes us immune to possible data races.

In this case however, it seems like ConnectionManager's async-ness is only there because of UserTokenStore needing a two-step initialization (`new` and `load()` later). Feels like we could easily avoid the problem by initializing both in the constructor, and hiding UTS' asyncness internally (since all its methods that require `load()` to complete first are `async` in the first place.

I'll draft us a way out of this, but maybe in another PR :)

Half-Shot: Yeah, promises are okay by me (if just a *little* ugly). And yeah, I loathe to have things which are possibly undefined after construction, so we should fix that.

src/AdminRoom.ts
```diff
@@ -423,6 +486,10 @@ export class AdminRoom extends AdminRoomCommandHandler {
         return this.botIntent.underlyingClient.sendStateEvent(this.roomId, NotifFilter.StateType, "", this.notifFilter.getStateContent());
     }
 
+    private canManageConnections(): boolean {
+        return this.config.checkPermissionAny(this.userId, BridgePermissionLevel.manageConnections);
```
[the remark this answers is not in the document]
tadzik: Hm, fair – at least for github specifically.

As for admin and modifying other people's rooms, is the idea that `manageConnections` allows you to manage *your own* connections while `admin` may also allow you to manage others'? In docs `admin` is currently annotated with "Currently, there are no admin features so this exists as a placeholder.", so I guess that'd be a first use for it and we could document it accordingly.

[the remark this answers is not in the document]
Half-Shot: Exactly yet. manageConnections is probably poorly named but allows you to create connections in rooms, but it doesn't let you take over other peoples. `admin` finally gets some usage here :)

PR #376, opened 2022-06-16
Add a .node-version file

jaywink:
Add it with 16. Helps people like me running avn for the right node version to be activated - https://github.com/wbyoung/avn


Half-Shot:
Hah, I hadn't heard of `avn`. I suppose it's an automatic version switcher for those running lots of different Node versions.

This is cool, though I wish it was part of the package.json file where we store our other version informations.

Half-Shot (approved):

PR #377, opened 2022-06-17
Fix oauth login urls for GitHub cloud having the wrong domain

### LINKED ISSUE #375: GitHub login provides a enterprise `api.github.com` url rather than a cloud URL

Half-Shot:
Fixes #375 

I ended up refactoring how we generate URLs, and adding some sanity tests.


tadzik (approved):
No complaints. Yay for the tests!

PR #385, opened 2022-06-27
Update webhooks.md to explain the usage of Markdown in the transformation function

Qrrbrbirlbel:
Since the Markdown to HTML conversion that is used after the transformation function is applied does not include paragraph tags (`<p>`) the docs should tell the user about it (which could shave a lot of debugging time).

Also, the `data` object is always included with the provided message and can be vaccess by viewing the source of the message. (This also helps with troubleshooting.)

### DIFF
--- a/docs/setup/webhooks.md
+++ b/docs/setup/webhooks.md
@@ -98,6 +98,7 @@ Scripts are executed syncronously and expect the `result` variable to be set.
 
 If the script contains errors or is otherwise unable to work, the bridge will send an error to the room. You can check the logs of the bridge
 for a more precise error.
+The contents of the JSON object are provided with the message as well and can be accessed by the “view source” function of your client.
 
 ### V2 API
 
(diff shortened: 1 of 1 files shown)


docs/setup/webhooks.md
```diff
@@ -113,6 +114,12 @@ The `v2` api expects an object to be returned from the `result` variable.
 }
 ```
 
+If no HTML value is provided, `plain` will be interpreted as markdown by
+[MarkdownIt.renderInline](https://markdown-it.github.io/markdown-it/#MarkdownIt.renderInline)
+which does not support paragraph tags (`<p>`).
```
Half-Shot, line 119: Doesn't this just mean it doesn't wrap it in p tags. I think if you had newlines in your string it would still paragraph it?

Half-Shot (changes requested):

docs/setup/webhooks.md
Qrrbrbirlbel: `.renderInline` does not convert two new lines `Foo\n\nBar` in `<p>Foo</p><p>Bar</b>` while `.render` does. One `\n` gets converted to `<br>`, `\n\n` doesn't, if I remember it correctly. This leads to `Foo\nBar\n\nFoobar` to be converted to `Foo<br>Bar Foobar` which gets rendered as

> Foo  
> Bar Foobar

which is the opposite of what you'd at least expect.

Half-Shot: Ah okay. I always assumed `renderInline` was `render` without the `<p>` wrapping. We could just make hookshot use `render` if you'd prefer?

Half-Shot: Which is what you said above in https://github.com/matrix-org/matrix-hookshot/pull/385#issuecomment-1167846005. Do you want to make the change yourself, should just be a case of fiddling with https://github.com/matrix-org/matrix-hookshot/blob/main/src/Connections/GenericHook.ts#L380 and probably updating one or two tests?

PR #402, opened 2022-07-11
Use HTML formatting for webhook admin notices

AndrewFerr:
Signed-off-by: Andrew Ferrazzutti <andrewf@element.io>


Half-Shot (approved):
Does this fix a particular issue?

PR #406, opened 2022-07-11
Allow the same GitLab repo to be bridged to multiple rooms

Half-Shot:
This was a mistake where we didn't correctly scope the filter to the same room.


AndrewFerr (approved):
LGTM - correct & concise

PR #407, opened 2022-07-11
add Grafana dashboard

### LINKED ISSUE #147: provide a (sample) grafana dashboard
which could then also be added to matrix-docker-ansible-deploy thusly: https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/5a8b17c1df268e7fc206ed72bc8c7a7d56626c1b/group_vars/matrix_servers#L2125

HarHarLinks:
Fixes #147

image
image
image

There isn't too much going on in the pushes section, in particular I don't have jira and figma set up and i'm not sure if there are metrics for generic hooks.

I'm not a pro Grafana user, but I think it's a solid start.

A possible improvement would be to limit the y axis legends to integers which I don't know how to do yet. I used templating to easily select the correct prometheus globally, however to my knowledge the use of templating prevents one from using Grafana alerts (for now).


Half-Shot:
Wow, this is really cool! It's certainly much more expanded than the one I use at home :)

hookshot-dashboard.json
```diff
@@ -0,0 +1,1689 @@
+{
```
Half-Shot, line 1: If I had to leave the tiniest of nits, could this go into a `contrib/` directory which is what Synapse do I think. Can you then link the file in the metrics documentation so people can find it?

HarHarLinks:
> Wow, this is really cool! It's certainly much more expanded than the one I use at home :)

Mayhaps you can check the parts I noted in OP against a bigger instance that uses those features?

docs/metrics.md
```diff
@@ -10,6 +10,8 @@ metrics:
   port: 9002
 ```
 
+An example dashboard that can be used with [Grafana](https://grafana.com) can be found at [/contrib/hookshot-dashboard.json](/contrib/hookshot-dashboard.json).
```
Half-Shot, line 13: That *may* totally break horribly on github pages. Can you do a direct link ?

```suggestion
An example dashboard that can be used with [Grafana](https://grafana.com) can be found at [/contrib/hookshot-dashboard.json](https://github.com/matrix-org/matrix-hookshot/blob/main/contrib/hookshot-dashboard.json).
```

Half-Shot (approved):
Looking great!

HarHarLinks:
This is giving me some trouble. It's using `rate(process_cpu_seconds_total[$__rate_interval])`.

Ever since I restarted the bridge, resetting the counter, I only get "No data". Even though the `rate()` function is supposed to handle such resets:

> Breaks in monotonicity (such as counter resets due to target restarts) are automatically adjusted for.

Similarly for the many panels using similar functions.

HarHarLinks:
I also don't understand what the CI issue is, so please advise if necessary.

HarHarLinks:
So the issue above appears related to how often my prometheus polls metrics vs the range the derivation functions ((i)rate, (i)delta, etc) are applied on. I'm still not sure why it would work before the first counter reset and not after.

I am able to work around this issue by telling Grafana these 2 global variables:
image
Interval is my sampling interval, 2x Interval is simply twice that, since I was unable to do this programmatically (it appears these don't support arithmetics).

docs/metrics.md
```diff
@@ -10,6 +10,13 @@ metrics:
   port: 9002
 ```
 
+An example dashboard that can be used with [Grafana](https://grafana.com) can be found at [/contrib/hookshot-dashboard.json](https://github.com/matrix-org/matrix-hookshot/blob/main/contrib/hookshot-dashboard.json).
```
Half-Shot, line 13: Sorry, this is non-obvious: You need to update https://github.com/matrix-org/matrix-hookshot/blob/main/scripts/build-metrics-docs.ts#L26, and then run `build:docs`.

HarHarLinks: Could you clarify the required steps in https://github.com/matrix-org/matrix-hookshot/blob/main/CONTRIBUTING.md ?
I can't simply `yarn build:docs` as I need to install requirements first etc and may not be familiar with that.

Half-Shot: Should be just `yarn` as mentioned in the setup instructions. The mdbook steps are hidden in the readme, but installing it is just `cargo install mdbook`.

We could probably have a page for developer setup instructions too.

HarHarLinks:
This actual data
image
when put through `idelta` results in this weird dip
image
instead of being thrown out as a reset and idk why it would do that.

Otherwise works well as war as I can see.

contrib/hookshot-dashboard.json
```diff
@@ -0,0 +1,2135 @@
...
+        "targets": [
+          {
+            "datasource": {
+              "type": "prometheus",
+              "uid": "${datasource}"
+            },
+            "editorMode": "builder",
+            "expr": "idelta(hookshot_connection_event_failed[$__interval])",
+            "interval": "$delta",
+            "legendFormat": "{{__name__}}",
```
Half-Shot, line 325: This should be this probably.

```suggestion
            "legendFormat": "{{ event }}",
```

Half-Shot (approved):
Have now tried it on EMS and it looks great, one nit!

contrib/hookshot-dashboard.json
HarHarLinks: One could use that as well as `connectionId`, I have not seen any such event to play around with. This panel is also pretty small and in the overview part at the top of the dashboard. My idea was to see at a glance if anything went wrong, however to use it to properly track down issues, one might need a bigger one to view more detail like `event` and `connectionId`.

So if you still say so, sure it's not unreasonable.

PR #410, opened 2022-07-12
Add XML parser for generic webhook payloads

Half-Shot:
Does what it says on the tin. We also now document what formats we support in the documentation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "uuid": "^8.3.2",
     "vm2": "^3.9.6",
     "winston": "^3.3.3",
+    "xml2js": "^0.4.23",
     "yaml": "^1.10.2"
   },
   "devDependencies": {
@@ -82,6 +83,7 @@
     "@types/node": "^14",
     "@types/node-emoji": "^1.8.1",
     "@types/uuid": "^8.3.3",
+    "@types/xml2js": "^0.4.11",
     "@typescript-eslint/eslint-plugin": "^5.4.0",
     "@typescript-eslint/parser": "^5.4.0",
     "@uiw/react-codemirror": "^4.5.3",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -62,6 +62,7 @@
     "uuid": "^8.3.2",
     "vm2": "^3.9.6",
     "winston": "^3.3.3",
+    "xml2js": "^0.4.23",
```
Half-Shot, line 65: I chose this because we already use it in the rss parser.

src/generic/Router.ts
```diff
@@ -55,13 +55,34 @@ export class GenericWebhooksRouter {
...
+        });
+    }
+
     public getRouter() {
         const router = Router();
         router.all(
             '/:hookId',
-            express.text({ type: 'text/*'}),
+            GenericWebhooksRouter.xmlHandler,
             express.urlencoded({ extended: false }),
```
AndrewFerr, line 83: Is this for `x-www-form-urlencoded`?  And whatever it is, is it a distinct enough type to be documented in the "Payload formats" section?

AndrewFerr (changes requested):

src/generic/Router.ts
Half-Shot: Good point. I think I missed it out.

src/provisioning/provisioner.ts
```diff
@@ -70,7 +70,6 @@ export class Provisioner {
             (...args) => this.checkUserPermission("write", ...args),
             this.deleteConnection.bind(this),
         );
-        this.expressRouter.use((err: unknown, req: Request, res: Response, next: NextFunction) => errorMiddleware(log)(err, req, res, next));
```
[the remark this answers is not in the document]
Half-Shot: Failed to do a Ctrl-S, oops.

src/generic/Router.ts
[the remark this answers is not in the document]
Half-Shot: Ah yes, most probably!

AndrewFerr (approved):

PR #417, opened 2022-07-15
Update matrix-bot-sdk to 0.6.0

Half-Shot:
Fixes https://github.com/matrix-org/matrix-hookshot/issues/435

This brings us up to date, and also unlocks the encryption support in a later PR. It also makes us require Node 16.

~~Blocked by requiring Node 16, and https://github.com/turt2live/matrix-bot-sdk/issues/247~~


package.json
```diff
@@ -53,7 +53,7 @@
     "jira-client": "^8.0.0",
     "markdown-it": "^12.3.2",
     "matrix-appservice-bridge": "^4.0.1",
-    "matrix-bot-sdk": "^0.5.19",
+    "matrix-bot-sdk": "^0.6.0",
```
AndrewFerr, line 56: Since https://github.com/turt2live/matrix-bot-sdk/issues/247 was fixed, this can use `^0.6.1` now, right?

Half-Shot: Done!

src/Connections/GitlabIssue.ts
```diff
@@ -44,8 +44,8 @@ export class GitLabIssueConnection extends BaseConnection implements IConnection
     static readonly QueryRoomRegex = /#gitlab_(.+)_(.+)_(\d+):.*/;
     static readonly ServiceCategory = "gitlab";
 
-    static getTopicString(authorName: string, state: string) {
```
tadzik, line 47: Huh, (how) did this work before? Do we even use this?

.github/workflows/main.yml
```diff
@@ -48,14 +48,14 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 16
-      - run: yarn --ignore-scripts
+      - run: yarn # Need to build to get rust bindings
```
tadzik, line 51:
```suggestion
      - run: yarn # Need to build scripts to get rust bindings
```

Just to make it more obvious :)

tadzik (approved):

src/Connections/GitlabIssue.ts
Half-Shot: Don't think this was ever used :disappointed:

tadzik: I guess it's time to drop it then :)

PR #418, opened 2022-07-15
Share code for tracking config via state events

AndrewFerr:
This should prevent forgetting to add state event handlers in new or
existing connection types.

### DIFF
--- a/src/Connections/CommandConnection.ts
+++ b/src/Connections/CommandConnection.ts
@@ -3,31 +3,40 @@ import LogWrapper from "../LogWrapper";
 import { MatrixClient } from "matrix-bot-sdk";
 import { MatrixMessageContent, MatrixEvent } from "../MatrixEvent";
 import { BaseConnection } from "./BaseConnection";
-import { PermissionCheckFn } from ".";
+import { IConnectionState, PermissionCheckFn } from ".";
 const log = new LogWrapper("CommandConnection");
 
 /**
  * Connection class that handles commands for a given connection. Should be used
  * by connections expecting to handle user input.
  */
-export abstract class CommandConnection extends BaseConnection {
+export abstract class CommandConnection<StateType extends IConnectionState = IConnectionState> extends BaseConnection {
     protected enabledHelpCategories?: string[];
     protected includeTitlesInHelp?: boolean;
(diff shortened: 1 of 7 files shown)


src/Connections/CommandConnection.ts
```diff
@@ -3,31 +3,40 @@ import LogWrapper from "../LogWrapper";
...
     protected get commandPrefix() {
-        return this.stateCommandPrefix + " ";
+        return (this.state.commandPrefix || this.defaultCommandPrefix) + " ";
+    }
+
+    public async onStateUpdate(stateEv: MatrixEvent<unknown>) {
+        this.state = this.validateConnectionState(stateEv.content);
+    }
+
+    protected validateConnectionState(content: unknown): StateType {
```
Half-Shot, line 38: I forget if you can mark a function as abstract, but maybe we should throw here if this isn't overridden. It's not great that we default to allowing any old content.

src/Connections/SetupConnection.ts
```diff
@@ -41,6 +41,8 @@ export class SetupConnection extends CommandConnection {
                 roomId,
                 "",
                 "",
+                // TODO Consider storing room-specific config in state.
```
Half-Shot, line 44: Not sure what to do here, a SetupConnection has no state.

AndrewFerr: True, but it may be useful for it to have state (like a custom command prefix).

The current situation with state is a bit wonky, though: the repo connection classes are stateful while SetupConnection is stateless, despite all of them being "command-controlled" while only the repo connections save their command prefix in state.  I couldn't think of a good way to share the state-setting code without it being inherited by SetupConnection as well (using mixins is an option, but is probably more trouble than it's worth), nor could I think of a way to separate out the code for handling command prefixes & saving state; so as it stands currently, SetupConnection holds an empty state object.

src/Connections/CommandConnection.ts
AndrewFerr: This is now done.

AndrewFerr:
Force-pushed to more cleanly merge this & the GitHub widget PR with the docstring fix.

PR #420, opened 2022-07-22
Add GitHub widget

AndrewFerr:
Also make assorted fixes to help push this along

---

Depends on https://github.com/matrix-org/matrix-hookshot/pull/418. Can be split into separate commits if desirable.

Open pre-existing issues (which affect other widgets too):
- Setting the command prefix via the widget doesn't work.
- CSS doesn't indent "child" event checkboxes (like Merge Requests->Opened/Closed/...), so they're all laid out in a single list despite having parent-child dependencies.


AndrewFerr:
Examples:
image
image

AndrewFerr:
Org names are now shown, which is needed to avoid conflicts between repos with the same name but in different orgs.

image

src/ConnectionManager.ts
```diff
@@ -293,15 +293,24 @@ export class ConnectionManager extends EventEmitter {
...
-                throw new ApiError('GitLab is not configured', ErrCode.DisabledFeature);
-            }
-            if (!this.config.checkPermission(userId, "gitlab", BridgePermissionLevel.manageConnections)) {
-                throw new ApiError('User is not permitted to provision connections for GitLab', ErrCode.ForbiddenUser);
-            }
-            return await GitLabRepoConnection.getConnectionTargets(userId, this.tokenStore, this.config.gitlab, filters);
+        switch (type) {
+        case GitLabRepoConnection.CanonicalEventType:
+            this.validateConnectionTarget(userId, this.config.gitlab, "GitLab", "gitlab");
+            return await GitLabRepoConnection.getConnectionTargets(userId, this.tokenStore, this.config.gitlab!, filters);
```
Half-Shot, line 299: No non-nulls, prefer to throw if it's not defined. (As we did before)

Half-Shot, line 299: No non-nulls, prefer to throw if it's not defined. (As we did before)

src/ConnectionManager.ts
```diff
@@ -293,15 +293,24 @@ export class ConnectionManager extends EventEmitter {
...
+        case GitHubRepoConnection.CanonicalEventType:
+            this.validateConnectionTarget(userId, this.config.github, "GitHub", "github");
+            return await GitHubRepoConnection.getConnectionTargets(userId, this.tokenStore, this.config.github!);
+        default:
+            throw new ApiError(`Connection type not known`, ErrCode.NotFound);
```
Half-Shot, line 304:
```suggestion
            throw new ApiError(`Connection type doesn't support getting targets or is not known`, ErrCode.NotFound);
```

src/ConnectionManager.ts
Half-Shot: Ah, I see that's what the above is doing. I'd still return the config or something from the validateConnectionTarget to avoid non-nulls, because it basically ignores the typing system which will bite us one day.

web/components/roomConfig/GenericWebhookConfig.tsx
```diff
@@ -40,7 +40,7 @@ const ConnectionConfiguration: FunctionComponent<ConnectionConfigurationProps<Se
             return;
         }
         onSave({
-            name: nameRef?.current?.value || existingConnection?.config.name,
+            name: nameRef?.current?.value || existingConnection?.config.name || "Generic Webhook",
```
Half-Shot, line 43: What issue is this fixing?

Half-Shot (changes requested):
On the right path, I just have a few questions

src/ConnectionManager.ts
AndrewFerr: Thanks for the suggestion.  This is addressed in the latest commit.

src/Connections/GithubRepo.ts
```diff
@@ -415,8 +425,8 @@ export class GitHubRepoConnection extends CommandConnection implements IConnecti
         return this.state.priority || super.priority;
     }
 
-    public async onStateUpdate(stateEv: MatrixEvent<unknown>) {
-        this.state = stateEv.content as GitHubRepoConnectionState;
+    protected validateConnectionState(content: unknown) {
+        return content as GitHubRepoConnectionState;
```
Half-Shot, line 429: Seems dangerous, can't we use the `validateState` function?

src/Connections/GithubRepo.ts
```diff
@@ -1042,6 +1052,32 @@ ${event.release.body}`;
         }
     }
 
+    public static async getConnectionTargets(userId: string, tokenStore: UserTokenStore, config: BridgeConfigGitHub): Promise<GitHubRepoConnectionTarget[]> {
+        // Search for all repos under the user's control.
+        const octokit = await tokenStore.getOctokitForUser(userId);
+        if (!octokit) {
+            throw new ApiError("User is not authenticated with GitHub", ErrCode.ForbiddenUser);
+        }
+        const allRepos = await octokit.repos.listForAuthenticatedUser({
```
Half-Shot, line 1061: Does this only return repos the user can actually control (i.e. create webhooks for), or all the repos the user can see?

src/Connections/GitlabRepo.ts
```diff
@@ -127,10 +126,10 @@ export interface GitLabTargetFilter {
 }
 
 /**
- * Handles rooms connected to a github repo.
+ * Handles rooms connected to a gitlab repo.
```
Half-Shot, line 129:
```suggestion
 * Handles rooms connected to a GitLab repo.
```

Half-Shot (changes requested):
A few nits, then it should be good to go!

src/Connections/GithubRepo.ts
AndrewFerr: We could, as long as there was no strong reason why it wasn't being done already.

src/Connections/GithubRepo.ts
AndrewFerr: The docstring says it's the latter:
> Lists repositories that the authenticated user has explicit permission (:read, :write, or :admin) to access.
> 
> The authenticated user has explicit permission to access repositories they own, repositories where they are a collaborator, and repositories that they can access through an organization membership.

Is write permission needed for adding read-only webhooks?

src/Connections/GitlabRepo.ts
AndrewFerr: Now that you mention it, there are a few of these that need updating, so I'll just make a separate PR for them.

src/Connections/GithubRepo.ts
AndrewFerr: 6b08b7b

src/Connections/GitlabRepo.ts
AndrewFerr: #428

src/Connections/GithubRepo.ts
Half-Shot: Surely you need at least write permissions, as `read` doesn't grant you the ability to edit the settings?

AndrewFerr:
One final (zero-diff) force-push to merge cleanly on the squashed version of #418.

PR #424, opened 2022-07-25
Fix typo: propetyKey -> propertyKey

jaller94:


Half-Shot (approved):
This probably doesn't need a changelog. Thanks!

PR #429, opened 2022-07-29
Skip already handled GitHub events

Half-Shot:
This is just a simple check to ensure we don't handle multiple events sent by GitHub by looking at the GUID provided. This is already pretty rare, but we've seen this happen rarely in the wild.

GitLab/JIRA don't support this as they don't provide a guid :s

### DIFF
--- a/src/Webhooks.ts
+++ b/src/Webhooks.ts
@@ -6,7 +6,7 @@ import LogWrapper from "./LogWrapper";
 import qs from "querystring";
 import axios from "axios";
 import { IGitLabWebhookEvent, IGitLabWebhookIssueStateEvent, IGitLabWebhookMREvent, IGitLabWebhookReleaseEvent } from "./Gitlab/WebhookTypes";
-import { EmitterWebhookEvent, Webhooks as OctokitWebhooks } from "@octokit/webhooks"
+import { EmitterWebhookEvent, EmitterWebhookEventName, Webhooks as OctokitWebhooks } from "@octokit/webhooks"
 import { IJiraWebhookEvent } from "./Jira/WebhookTypes";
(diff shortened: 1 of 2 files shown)


src/Webhooks.ts
```diff
@@ -139,15 +141,20 @@ export class Webhooks extends EventEmitter {
         try {
             let eventName: string|null = null;
             const body = req.body;
-            if (req.headers['x-hub-signature']) {
```
tadzik, line 142: We check for the a different header now, but Github [always sends both](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads).

src/Webhooks.ts
```diff
@@ -37,8 +38,9 @@ export interface NotificationsDisableEvent {
 export class Webhooks extends EventEmitter {
     
     public readonly expressRouter = Router();
-    private queue: MessageQueue;
-    private ghWebhooks?: OctokitWebhooks;
+    private readonly queue: MessageQueue;
+    private readonly ghWebhooks?: OctokitWebhooks;
+    private readonly guids = new QuickLRU<string, void>({ maxAge: 5000, maxSize: 100 });
```
tadzik, line 43: Maybe `handledGuids` (, to make it more obvious what we keep them for.

tadzik (changes requested):

src/Webhooks.ts
Half-Shot: I think this is okay, the delivery header is always sent?

tadzik: Yeah, that's what I meant :)

tadzik (approved):

PR #440, opened 2022-08-11
Validate figma access token + cleaner errors

Half-Shot:
After seeing vague errors in production, it felt wise to make our errors cleaner, and also check that the access token is valid.

### DIFF
--- a/src/figma/index.ts
+++ b/src/figma/index.ts
@@ -1,7 +1,7 @@
 import { BridgeConfigFigma } from "../Config/Config";
 import * as Figma from 'figma-js';
 import { MatrixClient } from "matrix-bot-sdk";
-import { AxiosError, isAxiosError } from "axios";
+import { AxiosError } from "axios";
 import LogWrapper from "../LogWrapper";
 
 export * from "./router";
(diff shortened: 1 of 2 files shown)


src/figma/index.ts
```diff
@@ -15,47 +16,77 @@ interface FigmaWebhookDefinition {
...
+                    team_id: teamId,
                 }, axiosConfig);
+                webhookDefinition = res.data as FigmaWebhookDefinition;
+                await matrixClient.setAccountData(accountDataKey, {webhookId: webhookDefinition.id});
+            } catch (ex) {
+                const axiosErr = ex as AxiosError;
+                if (axiosErr.isAxiosError) {
+                    log.error(`Failed to create webhook: ${axiosErr.code} ${axiosErr.response?.data?.message ?? ""}`)
+                }
+                throw Error(`Could not create a Figma webhook for instance ${instanceName}`);
```
tadzik, line 87:
```suggestion
                throw Error(`Could not create a Figma webhook for instance ${instanceName}: ${ex}`);
```

tadzik (changes requested):

tadzik (approved):

PR #441, opened 2022-08-11
Add query parameter to scope room config widget to a particular service

### LINKED ISSUE #437: Add ability to automatically select a service type when opening a room config widget
I.e. a query parameter like `defaultService=jira`

justinbot:
Adds support for a `serviceScope` query parameter in the config widget URL, which makes it go directly to a particular service (and removes navigation back to the list of services).

For example with `serviceScope=feeds`:

Screen Shot 2022-08-11 at 2 04 22 PM

If the value provided for `serviceScope` does not correspond to a supported service, behavior will be as if it wasn't provided.

Fixes #437

### DIFF
--- a/changelog.d/ 441.feature
+++ b/changelog.d/ 441.feature
@@ -1,0 +1,1 @@
+Add query parameter to scope room config widget to a particular service.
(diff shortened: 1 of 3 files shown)


Half-Shot (approved):
No complaints from me!

Issue #456, opened 2022-08-24
Provisioning fails if bot isn't already in room


AndrewFerr:
**Describe the bug**
When using a Hookshot-equipped Dimension to check available bridges for a room that the Hookshot bot isn't already a member of, Dimension shows Hookshot bridges (Github, Webhooks, etc) as unavailable.

This happens because this flow triggers a bad property lookup in Hookshot code.

**To Reproduce**
1. Configure Dimension & Hookshot
2. Create a new room
3. In Element Web, click ⓘ -> "Add widgets, bridges & bots"
4. See Hookshot bridges displayed with a ⚠
5. Click on any of the bridges to see an error popup of "This integration is offline or unavailable"

This can be fixed by inviting the Hookshot bot to the room.

**Expected behavior**
Dimension's bridge list appear to hit errors if Hookshot's bot isn't in a room.

If Hookshot bridges require the bot to be in the room, it should ideally be auto-invited.

**Screenshots**
![image](https://user-images.githubusercontent.com/3271094/186507711-3ba4a99b-9cc2-47e7-a49c-655ced0a5014.png)

Half-Shot:
#456 fixed this.

PR #457, opened 2022-08-24
Don't look for `body` in an ApiError

AndrewFerr:
Instead, look for `jsonBody`

### DIFF
--- a/changelog.d/457.bugfix
+++ b/changelog.d/457.bugfix
@@ -1,0 +1,1 @@
+Fix error when responding to a provisioning request for a room that the Hookshot bot isn't yet a member of.
--- a/src/provisioning/api.ts
+++ b/src/provisioning/api.ts
@@ -27,7 +27,7 @@ export async function assertUserPermissionsInRoom(userId: string, roomId: string
             throw new ApiError("Bot is not joined to the room.", ErrCode.NotInRoom);
         }
     } catch (ex) {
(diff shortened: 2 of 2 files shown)


AndrewFerr:
This could have been more easily noticed if [type-checking on errors wasn't disabled](https://github.com/matrix-org/matrix-hookshot/blob/563603ff76be724fb7f2531a291813ea56ade169/tsconfig.json#L26-L27) (thanks @justinbot for pointing out that config!).

It's still too early to turn it on, though, as Hookshot still uses a version of `matrix-bot-sdk` that doesn't expose its ubiquitous `MatrixError` type, which would be one of the primary errors to type-check.

AndrewFerr:
Oh, and this should fix https://github.com/matrix-org/matrix-hookshot/issues/456

AndrewFerr:
Just confirmed that this indeed fixes #456:
image

Note that I'm the only person in the room (see "People 1" on the right sidebar) and no error markers appear. The "Bridges" items are clickable & work properly.

Half-Shot (approved):
Looks good. I've patched upstream so we can access the error object https://github.com/turt2live/matrix-bot-sdk/pull/257.

Technically you can still access the error object by importing the MatrixError.ts file directly in the code, but I wouldn't begrudge you for wanting to do something less hacky!

PR #392, opened 2022-06-30
Revert "Add docker auto-build for ARM64"

Half-Shot:
Reverts matrix-org/matrix-hookshot#373

Fixes #390. Sadly https://github.com/matrix-org/matrix-rust-sdk-bindings does not compile successfully on arm64. Cargo runs, but is killed (assumed to be resource pressure) when running in GitHub actions. It looks to be a problem for arm64 because no binarys are packaged for its architecture in the npm package.

The up and coming refactor in matrix-bot-sdk to use the rust-sdk directly might mean we can re-enable arm64 builds again, but at this moment in time it's simply too much of a time sink to work out why a binding layer we do not use (but, because the bot-sdk makes it a required dependency, we have to build it anyway) fails to compile.

PR #458, opened 2022-08-24
Add docker auto-build for ARM64 (take 2)

Half-Shot:
Reverts matrix-org/matrix-hookshot#392

### DIFF
--- a/.github/workflows/docker-hub-latest.yml
+++ b/.github/workflows/docker-hub-latest.yml
@@ -7,7 +7,7 @@ on:
 
 env:
   DOCKER_NAMESPACE: halfshot
-  PLATFORMS: linux/amd64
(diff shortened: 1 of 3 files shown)


justinbot (approved):
LGTM!
However not totally clear why this works now but didn't before? Was it the update to matrix-bot-sdk?

Half-Shot:
> However not totally clear why this works now but didn't before? Was it the update to matrix-bot-sdk?

Yep. So we moved away from @turt2live/matrix-sdk-crypto-nodejs to https://github.com/matrix-org/matrix-rust-sdk/tree/main/bindings/matrix-sdk-crypto-nodejs. The old package didn't have arm64 prebuilts so we just failed to compile them, but the new package has prebuilts so it just works :)

PR #459, opened 2022-08-26
Refactor feed errors and error presentation

### LINKED ISSUE #382: RSS feed reader sends a warning to the room if it errors with "socket hangup"
These are temporary errors that aught to be retried (and probably, we should just skip the run silently and wait until the next cycle). It sounds like at the moment we default to failing loudly unless we know the type of failure, I wonder whether it would be better to fail silently unless the error is "fatal". (i.e. 401/403/ bad host etc).

Half-Shot:
Fixes #382 

This PR turned into a bit of a general feeds & errors refactor, so I would suggest to review commit by commit.

In terms of the overall changes, we now:
- Are silent on the *first* occurance of a non-serious error, like a connection reset.
- Show recent successes and failures in the widget, if you are an admin of the room.
- Refactored the errors reported by the validator of the Feed connection, so they read a bit better.
- Added the ability to set a timeout for feeds, so that we don't wait forever for one feed to be read. Defaults to 10 seconds.
- Importantly, errors from the API are no longer hidden and are visible to the user. We've also tarted up the ErrorPane a bit to look nicer.

This looks like:

image

In terms of the error reporter:

image


src/Config/Config.ts
```diff
@@ -232,15 +232,20 @@ export class BridgeConfigGitLab {
...
     public enabled: boolean;
     public pollIntervalSeconds: number;
+    public pollTimeoutSeconds: number;
 
     constructor(yaml: BridgeConfigFeedsYAML) {
         this.enabled = yaml.enabled;
         this.pollIntervalSeconds = yaml.pollIntervalSeconds;
+        assert.strictEqual(typeof this.pollIntervalSeconds, "number");
+        this.pollTimeoutSeconds = yaml.pollTimeoutSeconds ?? 10;
+        assert.strictEqual(typeof this.pollTimeoutSeconds, "number");
```
tadzik, line 248: Duplicate assert, I don't think we need both :)

src/Connections/FeedConnection.ts
```diff
@@ -10,16 +10,32 @@ import axios from "axios";
...
+}
+
+
 export interface FeedConnectionState extends IConnectionState {
     url:    string;
     label?: string;
 }
 
-export type FeedResponseItem = GetConnectionsResponseItem<FeedConnectionState, object>;
+export interface FeedConnectionSecrets {
```
tadzik, line 34: Why is that part of secrets?

src/Connections/FeedConnection.ts
```diff
@@ -37,14 +53,19 @@ export class FeedConnection extends BaseConnection implements IConnection {
...
+        let res;
+        try {
+            res = await axios.head(url).catch(() => axios.get(url));
+        } catch (ex) {
+            throw new ApiError(`Could not read from URL: ${ex.message}`, ErrCode.BadValue);
+        }
+        const contentType = res.headers['content-type'];
+        // we're deliberately liberal here, since different things pop up in the wild
+        if (!contentType.match(/xml/)) {
+            throw new ApiError(`Feed responded with a content type of "${contentType}", which doesn't look like an RSS/Atom feed`);
```
tadzik, line 68: Could include something like 415 (Unsupported media type) in the api error.

src/Connections/FeedConnection.ts
```diff
@@ -95,10 +112,16 @@ export class FeedConnection extends BaseConnection implements IConnection {
                 url: this.feedUrl,
                 label: this.state.label,
             },
+            secrets: {
+                lastResults: [...this.lastResults.entriesDescending()].map(e => e[1]),
```
tadzik, line 116: I'm not sure about this use of QuickLRU as a circular buffer :P I guess it does the job, but I think it brings a bit more mental overhead than an array with an index tracker (or just array we push to, and shorten in at the front) would.

src/Connections/FeedConnection.ts
```diff
@@ -95,10 +112,16 @@ export class FeedConnection extends BaseConnection implements IConnection {
...
+            }
         }
     }
 
     private hasError = false;
+    private readonly lastResults = new QuickLRU<string, LastResultOk|LastResultFail>({ maxSize: 5 });
+    // Simply for fast lookup.
+    private lastResult: FeedError|true = true;
```
tadzik, line 124: I'd rather have this be a getter extracting things from lastResults, so that we don't have to set it twice.

src/Connections/FeedConnection.ts
tadzik: (Or just a two-element array, if we only look back once).

tadzik (changes requested):
Looking good! I think the QuickLRU is a bit of an overkill (and more trouble than it's worth in our case) though – how about we dumb it down a bit?

src/Config/Config.ts
Half-Shot: I don't think this is duped?

src/Connections/FeedConnection.ts
Half-Shot: I didn't want to expose the status of the feed to everyone in the room, but just room admins. We can rethink that logic though, wdyt?

src/Connections/FeedConnection.ts
Half-Shot: Done in [6cd3840](https://github.com/matrix-org/matrix-hookshot/pull/459/commits/6cd3840c6104b737c87497f2cca385bfa8456612)

src/Connections/FeedConnection.ts
Half-Shot: Yeah, fair point :)

src/Config/Config.ts
tadzik: Oh yeah, I can't read.

src/Connections/FeedConnection.ts
tadzik: Oh, I thought of secrets as "visible to no one", but if they're admin-visible then that makes sense.

src/Connections/FeedConnection.ts
Half-Shot: Done in [5fe3d4e](https://github.com/matrix-org/matrix-hookshot/pull/459/commits/5fe3d4e3d69d9ee361d4c7dec8e2f00a195e16c7)

src/Connections/FeedConnection.ts
Half-Shot: Done in [5fe3d4e](https://github.com/matrix-org/matrix-hookshot/pull/459/commits/5fe3d4e3d69d9ee361d4c7dec8e2f00a195e16c7)

Half-Shot:
And that's what I get for trying to rebase in a version change from `origin/main` -_-

src/Connections/FeedConnection.ts
tadzik: (Actually 9e61ff7d8919bec97f8a4f392b5f5fe288f6f0fd post-rebase, it seems)

tadzik: Looks alright!

PR #462, opened 2022-08-31
Update ioredis & block on startup if we couldn't connect

Half-Shot:
Our ioredis was a touch of of date and so I've updated it to 5.x. Doesn't seem to have broken any behaviours for us, and we now have native types and native promises \o/.

We also now block startup until Redis is ready, to avoid us trying to handle traffic without a working storage layer.


tadzik (approved):
Looks sensible :+1:

PR #466, opened 2022-09-01
Add tests for generic webhook routing

Half-Shot:
This PR tests that all the webhook payload types are handled, and the error cases.

I wanted to use an express mock, but it turned out to be a lot of faff on Mocha. For the time being we're using using a live express instance since it then ensures we're testing the real thing.


src/generic/Router.ts
```diff
@@ -41,13 +42,13 @@ export class GenericWebhooksRouter {
                     next();
                     return;
                 }
-                res.status(404).send({ok: false, error: "Webhook not found"});
+                res.status(StatusCodes.NOT_FOUND).send({ok: false, error: "Webhook not found"});
```
AndrewFerr, line 45: Nice -- and let's make a follow-up change to do this for the other Routers

AndrewFerr (approved):
LGTM

PR #467, opened 2022-09-02
Logging fixes

Half-Shot:
#463 broke logging so that everything now logs at debug, which is less than helpful.

This also fixes the case where we don't log out errors before logging is configured.

### DIFF
--- a/src/LogWrapper.ts
+++ b/src/LogWrapper.ts
@@ -161,7 +161,7 @@ export default class LogWrapper {
      * @param {*[]} messageOrObject The data to log
      */
     public debug(...messageOrObject: MsgType[]) {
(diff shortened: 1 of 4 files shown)


src/LogWrapper.ts
```diff
@@ -29,6 +29,12 @@ interface HookshotLogInfo extends winston.Logform.TransformableInfo {
 }
 export default class LogWrapper {
 
+    static isConfigured: boolean;
```
AndrewFerr, line 32: Should this be `private`, since it has a getter?

Half-Shot: Yes -- this is not rust

AndrewFerr (approved):

PR #469, opened 2022-09-02
Fix feed.success crashing the process due to a typing issue

Half-Shot:
This caused:

```
/bin/matrix-hookshot/Bridge.js:402
        this.bindHandlerToQueue("feed.success", (data) => connManager.getConnectionsForFeedUrl(data.feed.url), c => c.handleFeedSuccess());
                                                                                                    ^

TypeError: Cannot read properties of undefined (reading 'feed')
    at Bridge.<anonymous> (/bin/matrix-hookshot/Bridge.js:402:101)
    at LocalMQ.<anonymous> (/bin/matrix-hookshot/Bridge.js:494:61)
    at LocalMQ.emit (node:events:513:28)
    at LocalMQ.push (/bin/matrix-hookshot/MessageQueue/LocalMQ.js:31:14)
    at FeedReader.pollFeeds (/bin/matrix-hookshot/feeds/FeedReader.js:199:28)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)

```

### DIFF
--- a/src/Bridge.ts
+++ b/src/Bridge.ts
@@ -615,9 +615,9 @@ export class Bridge {
             (data) => connManager.getConnectionsForFeedUrl(data.feed.url),
             (c, data) => c.handleFeedEntry(data),
         );
-        this.bindHandlerToQueue<FeedEntry, FeedConnection>(
+        this.bindHandlerToQueue<{url: string}, FeedConnection>(
             "feed.success",
-            (data) => connManager.getConnectionsForFeedUrl(data.feed.url),
+            (data) => connManager.getConnectionsForFeedUrl(data.url),
             c => c.handleFeedSuccess(),
         );
         this.bindHandlerToQueue<FeedError, FeedConnection>(
--- a/src/Bridge.ts
+++ b/src/Bridge.ts
@@ -40,7 +40,7 @@ import { getAppservice } from "./appservice";
 import { JiraOAuthRequestCloud, JiraOAuthRequestOnPrem, JiraOAuthRequestResult } from "./Jira/OAuth";
 import { GenericWebhookEvent, GenericWebhookEventResult } from "./generic/types";
 import { SetupWidget } from "./Widgets/SetupWidget";
-import { FeedEntry, FeedError, FeedReader } from "./feeds/FeedReader";
+import { FeedEntry, FeedError, FeedReader, FeedSuccess } from "./feeds/FeedReader";
 const log = new LogWrapper("Bridge");
 
 export class Bridge {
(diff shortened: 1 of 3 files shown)


src/feeds/FeedReader.ts
```diff
@@ -210,7 +210,7 @@ export class FeedReader {
                     const entry = {
                         feed: {
                             title: feed.title ? stripHtml(feed.title) : null,
-                            url: url.toString()
```
AndrewFerr, line 213: TIL strings have `toString` :thinking: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toString

src/Bridge.ts
```diff
@@ -615,9 +615,9 @@ export class Bridge {
             (data) => connManager.getConnectionsForFeedUrl(data.feed.url),
             (c, data) => c.handleFeedEntry(data),
         );
-        this.bindHandlerToQueue<FeedEntry, FeedConnection>(
+        this.bindHandlerToQueue<{url: string}, FeedConnection>(
```
AndrewFerr, line 618: For future-proofing, how about defining this as an interface, like `FeedSuccess`?  I have it written ready if interested; not a blocker if a release is time-critical.

AndrewFerr (approved):

src/Bridge.ts
Half-Shot: Yeah, I think we should. Uno momento.

src/feeds/FeedReader.ts
Half-Shot: Yeah, I suspect previous versions of this code used `URL` which is why this lingered.

src/Bridge.ts
```diff
@@ -615,9 +615,9 @@ export class Bridge {
             (data) => connManager.getConnectionsForFeedUrl(data.feed.url),
             (c, data) => c.handleFeedEntry(data),
         );
-        this.bindHandlerToQueue<FeedEntry, FeedConnection>(
+        this.bindHandlerToQueue<FeedSuccess, FeedConnection>(
```
AndrewFerr, line 618: The import for this is missing?

src/feeds/FeedReader.ts
```diff
@@ -55,6 +55,10 @@ export interface FeedEntry {
     fetchKey: string,
 }
 
+export interface FeedSuccess {
+    url: string;
```
AndrewFerr, line 59: Nitpick, should this end with a colon instead?

AndrewFerr (changes requested):

src/feeds/FeedReader.ts
Half-Shot: comma? probably.

src/Bridge.ts
Half-Shot: Missed a Ctrl-S :s

AndrewFerr (approved):

PR #472, opened 2022-09-02
Add function to load generic webhook scripts from other rooms

Half-Shot:
**Obviously, going between the boundaries of rooms is quite scary so this will require an additional config flag to be set**

Currently webhooks have the issue where if you want to create multiple webhook connections with the same handler script, you need to copy and paste that script across rooms. This is quite tedious and prone to breakage.

To solve this problem, this PR adds a new function to the script executing environment called `loadMatrixScript`. By giving the function a roomId/alias and an eventId, it will attempt to load the script from another room and execute it in a separate context but including the data from the webhook.

This allows you to potentially reuse and chain together several scripts.

Obvously this is prone to all sorts of abuse like cyclic script calling and recursion. For now the system only allows you to call one script, though we might evolve that if the need arises.

```js
// Original script
if (data.counter === undefined) {
    result = {
        empty: true,
        version: "v2"
    };
  } else if (data.counter > data.maxValue) {
    result = {
          plain: `**Oh no!** The counter has gone over by ${data.counter - data.maxValue}`,
          version: "v2"
    };
  } else {
    result = {
          plain: `*Everything is fine*, the counter is under by${data.maxValue - data.counter}`,
          version: "v2"
    };
  }
```

```js
// Calling script
result = await loadMatrixScript("matrix:r/webhook-script-source:beefy/$dDzl3ho6l3JOAItxK2he_wwhW8I7tfvCixhU2Im0JLU");
```

### DIFF
--- a/src/Connections/GenericHook.ts
+++ b/src/Connections/GenericHook.ts
@@ -36,6 +36,8 @@ export interface GenericHookSecrets {
 
 export type GenericHookResponseItem = GetConnectionsResponseItem<GenericHookConnectionState, GenericHookSecrets>;
 
+const MatrixUriEventRegex = /^matrix:(roomid|r)\/(.+:.+)\/[a-zA-Z0-9+/]+/;
+
 /** */
 export interface GenericHookAccountData {
     /**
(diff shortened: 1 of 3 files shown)


Twi1ightSparkle:
Will I be able to manage scripts in my library room with the widget?

src/Connections/GenericHook.ts
```diff
@@ -292,8 +301,30 @@ export class GenericHookConnection extends BaseConnection implements IConnection
...
+        if (!matrixUri) {
+            throw Error('Not a valid matrix path. Use the event URI scheme from https://spec.matrix.org/v1.3/appendices/#matrix-uri-scheme');
+        }
+        const [ _, type, prefixlessRoomId, eventId ] = matrixUri;
+        let roomId = "!" + prefixlessRoomId;
+        if (type === "r") {
+            // RoomAlias -> resolve
+            roomId = await this.as.botClient.resolveRoom("#" + prefixlessRoomId);
+        }
+        const eventData = (await this.as.botClient.getEvent(roomId, eventId)).content as GenericHookConnectionState;
```
jaller94, line 319: I think `eventData` can be `undefined`, which will cause the next line to crash.

tests/connections/GenericHookTest.ts
```diff
@@ -247,6 +247,28 @@ describe("GenericHookConnection", () => {
             an_array_of: ["1.2345", "6.789"],
             floats: true,
         });
-
+    });
+    it("will import a script from another room", async () => {
+        const webhookData = {question: 'What is the meaning of life?', answer: 42};
+        const [connection, mq] = createGenericHook({name: 'test', transformationFunction: `const f = loadMatrixScript("matrix://!foo:bar/$baz:bar").then(f) f();`}, {
```
jaller94, line 253:
```suggestion
        const [connection, mq] = createGenericHook({name: 'test', transformationFunction: `const f = loadMatrixScript("matrix:roomid/foo:bar/e/baz:bar").then(f) f();`}, {
```

jaller94 (changes requested):
Incorrect parsing of `matrix:` URIs.

PR #474, opened 2022-09-05
Add `getMediaUrl` function for generic webhooks

### LINKED ISSUE #470: Add ability to handle image URLs from generic webhooks
**Is your feature request related to a problem? Please describe.**
If the JSON payload from a generic webhook includes a URL for an image, there currently is no simple way to include that image within a message. 

**Describe the solution you'd like**
Matrix-hookshot could convert the URL to MXC, perhaps via some transformation JavaScript function. Then a user could include that MXC address within the HTML or Markdown for their message. 

**Describe alternatives you've considered**
It's possible to simply include the URL in the message and if URL previews are turned on, the image will display at the bottom of the message. This isn't ideal though since not all homeservers or rooms have URL previews on. Also, the preview image is often cropped in odd ways, and the image can't be displayed within the body of the message.

Half-Shot:
Fixes #470 

This function will return a deferred URI, and then attempt to perform a download+upload to a media repo after the script has finished processing.

### DIFF
--- a/docs/setup/webhooks.md
+++ b/docs/setup/webhooks.md
@@ -177,3 +177,28 @@ if (data.counter > data.maxValue) {
     result = `*Everything is fine*, the counter is under by ${data.maxValue - data.counter}`
 }
 ```
+
+
+### Functions
+
+The scripting environment exposes some convienence functions for doing advanced transformations
+of your data. 
+
+#### `getMediaUrl(url: string)`
+
+This function will take an input URL and return a deferred MXC URL, which can then be inserted
+into your response to display an image. If the input URL fails to be resolved, the deferred
+URL will be left in place.
+
+There is a maximum timeout for the fetch request of 5 seconds.
+
+#### Example
+
+```js
+const mxc = getMediaUrl(data.myimage);
+return {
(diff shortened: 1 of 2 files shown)


docs/setup/webhooks.md
```diff
@@ -177,3 +177,28 @@ if (data.counter > data.maxValue) {
...
+There is a maximum timeout for the fetch request of 5 seconds.
+
+#### Example
+
+```js
+const mxc = getMediaUrl(data.myimage);
+return {
+  "version": "v2" // The version of the schema being returned from the function. This is always "v2".
+  "plain": `![My image](${mxc})`, // The plaintext value to be used for the Matrix message.
+  "html": `<img src="${mxc}" alt="My image"></img>`, // The HTML value to be used for the Matrix message. If not provided, plain will be interpreted as markdown.
```
HarHarLinks, line 202: I know that inline images are technically specced, but not so widely supported. Did you decide to not offer to send one as m.image while still being able to send another m.text message for another reason than splitting one webhook into 2 messages?

Half-Shot: Yeah, it's intentional. At this stage I'd rather not have one hook send a nearly infinite number of messages. I'd rather clients added support for this :)

PR #476, opened 2022-09-05
Add support for specifying `appservice` as a listener

Half-Shot:
This code is slightly janky, but allows us to specify the appservice listener alongside our other listeners. It does this by mocking out the express app slightly so that bot-sdk cannot start it, but reusing the route configuration with the ListenerService logic we use across the rest of hookshot.


Half-Shot:
We're not happy with the solution because it's obviously munging the Server object and hoping the bot-sdk doesn't change it's usage. Currently trying to explore proxied unix sockets as a solution.

PR #478, opened 2022-09-06
Add Logger tests

### LINKED ISSUE #477: JSON logger causes actions to fail when hsData.shift() fails

Half-Shot:
Fixes #477 

In addition to fixing the issue with JSON causing the bridge to error, this also adds some torture tests for the logger. It was a bit more refactor-y than I'd like but I think this does a good job of ensuring we won't break loggers so easily in the future.

### DIFF
--- a/src/LogWrapper.ts
+++ b/src/LogWrapper.ts
@@ -147,7 +147,7 @@ export class GlobalLogger {
         });
 
         LogService.setLevel(LogLevel.fromString(cfg.level));
-        //LogService.debug("LogWrapper", "Reconfigured logging");
+        LogService.debug("LogWrapper", "Reconfigured logging");
         this.isConfigured = true;
     }
 }
--- a/src/LogWrapper.ts
+++ b/src/LogWrapper.ts
@@ -47,7 +47,7 @@ export default class LogWrapper {
 
     static messageFormatter(info: HookshotLogInfo): string {
         const logPrefix = `${info.level} ${info.timestamp} [${info.module}] `;
-        return logPrefix + this.formatMsgTypeArray(info.data);
+        return logPrefix + this.formatMsgTypeArray(info.data ?? []);
     }
 
     static winstonLog: winston.Logger;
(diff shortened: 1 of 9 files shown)


src/LogWrapper.ts
```diff
@@ -27,36 +27,26 @@ function isMessageNoise(messageOrObject: MsgType[]) {
 interface HookshotLogInfo extends winston.Logform.TransformableInfo {
     data: MsgType[];
 }
-export default class LogWrapper {
-
-    private static isConfigured: boolean;
+export class GlobalLogger {
```
AndrewFerr, line 30: Is there such thing as a module-private constructor in TS/JS?  If nothing outside `LogWrapper.ts` needs to / should build a child logger, the visibility of `GlobalLogger`'s constructor could be restricted to just this file.  Of course there's no harm in keeping this exposed; was mostly just wondering if such scoping exists.

src/LogWrapper.ts
```diff
@@ -27,36 +27,26 @@ function isMessageNoise(messageOrObject: MsgType[]) {
 interface HookshotLogInfo extends winston.Logform.TransformableInfo {
     data: MsgType[];
```
AndrewFerr, line 28: `MsgType` is a union that includes `any`, which is what allowed this bug to slip through in the first place.

Is there a type that resolves to "any non-nullish values"?  Of course, the root of the problem is that `TransformableInfo.data` is `any`, so this might not help much...

AndrewFerr (approved):
Some minor optional tweaks requested, but looks good overall :+1:

src/LogWrapper.ts
Half-Shot: We're using it for the test sadly :(

src/LogWrapper.ts
Half-Shot: I mean really we're expecting to handle `any`, as any value should be loggable. I should probably just alias `MsgType` as `any`.

PR #481, opened 2022-09-12
Fix figma not working on startup

Half-Shot:
Because we decided to stick `v2/` in front of our client calls, all the requests would fail with a 404.


tadzik (approved):
Looks okay :+1:

PR #483, opened 2022-09-12
Increase feed.pollTimeoutSeconds from 10s to 30s

Half-Shot:
10s seems too low in practice for things like GitHub, and it does seem to be doing more harm than good.

### DIFF
--- a/changelog.d/483.misc
+++ b/changelog.d/483.misc
@@ -1,0 +1,1 @@
+Increase default `feeds.pollTimeoutSeconds` from 10 seconds to 30.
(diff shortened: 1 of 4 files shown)


tadzik:
We're testing this internally first to see if it has the desired effect.

HarHarLinks:
I've bumped my setting to 20s based on discussion in the hookshot matrix room and had no more feed timeout issues for 5 days when there used to be multiple per day.

Half-Shot:
Alrighty, in that case I think this is probably fine.

PR #487, opened 2022-09-14
Add ability to view request results / logs for generic webhooks

### LINKED ISSUE #479: Show a list of received requests in the widget UI for generic webhooks
Bonus points would be a log of the transformation function, if it failed

Half-Shot:
Fixes #479 

This works much like the RSS feed result system, except for hooks. This additionally stores logs, errors and a couple of headers to help script writers diagnose any issues with their scripts.

TODO: Make this actually pleasant to use and look at.

image

### DIFF
--- a/src/Bridge.ts
+++ b/src/Bridge.ts
@@ -564,12 +564,12 @@ export class Bridge {
                 try {
                     // TODO: Support webhook responses to more than one room
                     if (index !== 0) {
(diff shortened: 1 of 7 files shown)


Half-Shot:
@nadonomy could I get a timeline for review on this one?

Half-Shot:
I've been given some design pointers from Aaron, so I'm going to split those out to a second PR first.

PR #488, opened 2022-09-15
Use matrix-appservice-bridge logging

Half-Shot:
Counterpart to https://github.com/matrix-org/matrix-appservice-bridge/pull/412


package.json
```diff
@@ -80,7 +80,7 @@
     "@types/ajv": "^1.0.0",
     "@types/chai": "^4.2.22",
     "@types/cors": "^2.8.12",
-    "@types/express": "^4.17.13",
+    "@types/express": "4.17.13",
```
jaller94, line 83: What's so glorious about this particular version? 🤔

src/figma/index.ts
```diff
@@ -15,8 +17,8 @@ interface FigmaWebhookDefinition {
     description: string;
 }
 
-const log = new LogWrapper('FigmaWebhooks');
-
+const log = new Logger('FigmaWebhooks');
+
```
jaller94, line 21: Space character in an otherwise empty line.
But no harm done.

jaller94 (approved):
Looks good to me.

package.json
Half-Shot: I think it was typing conflicts with m-a-b. Tedious. I'll see if I can stop pinning.

PR #493, opened 2022-09-22
Increase timeout to 30s for GitLab MR comment / approval message merges

Half-Shot:

### DIFF
--- a/changelog.d/493.bugfix
+++ b/changelog.d/493.bugfix
@@ -1,0 +1,1 @@
+Fix GitLab approval/review-comments appearing as two messages when sent close together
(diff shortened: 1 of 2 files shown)


tadzik:
Feels a bit similar to https://github.com/matrix-org/matrix-hookshot/pull/483 – could we test it first on the environment where 5s is not enough to see if this'd indeed help?

HarHarLinks:
alternatively to showing the rss error on first timeout(?) and increasing the timeout from 10s to 30s, it could only show it on consecutive poll timeouts?
so if it times out once then it fails silently, but if $interval minutes later it times out again, then it sends an error to matrix.
the argument being that feeds should not be so time sensitive (it's polling tech not pushing tech) since you probably don't expect them to fulfill some realtime requirement.

another method would be to retry again after a timeout sooner than the default interval, e.g. 1 minute.

PR #502, opened 2022-09-30
Add Jira widget

AndrewFerr:
image

image


Half-Shot (approved):
Can't fault it, well done!

AndrewFerr:
Not logged in:
image

Logged in, but with 0 available instances (simulated by forcing an empty instance list in code):
image

PR #503, opened 2022-09-30
Add commands to remove/list Jira connections

AndrewFerr:
This adds bot-command equivalents of the list & remove features that are available in the provisioning API, and soon, the configuration widget (note that this PR is not required by / dependent on the widget PR).

Note that this stops lower-casing the "safe URL" because that URL gets used as a state event key, which is case-sensitive, and wasn't lower-cased by the provisioning API (and isn't by the config widget either) -- meaning a connection added via a bot command couldn't be removed by the provisioning API (or widget), and vice-versa!

I'm pointing this out since the name of "safe URL" suggests that the lower-casing was done for a good reason, but AFAICT it only makes things worse.  (The rest of the "safety" in pulling apart irrelevant parts of the URL works great, though.)

### DIFF
--- a/src/Connections/SetupConnection.ts
+++ b/src/Connections/SetupConnection.ts
@@ -1,7 +1,7 @@
 // We need to instantiate some functions which are not directly called, which confuses typescript.
 import { BotCommands, botCommand, compileBotCommands, HelpFunction } from "../BotCommands";
 import { CommandConnection } from "./CommandConnection";
-import { GenericHookConnection, GitHubRepoConnection, JiraProjectConnection } from ".";
+import { GenericHookConnection, GitHubRepoConnection, JiraProjectConnection, JiraProjectConnectionState } from ".";
 import { CommandError } from "../errors";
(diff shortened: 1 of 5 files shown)


src/Connections/SetupConnection.ts
```diff
@@ -1,7 +1,7 @@
 // We need to instantiate some functions which are not directly called, which confuses typescript.
```
Half-Shot, line 1: Could I get you to write a page similar to https://matrix-org.github.io/matrix-hookshot/latest/usage/room_configuration/github_repo.html so we have a record of how to setup a Jira connection via commands?

Half-Shot (changes requested):
Code looks good, but I'd love a document :)

src/Connections/SetupConnection.ts
AndrewFerr: Done!

Half-Shot (approved):

PR #505, opened 2022-09-30
Make various typechecking improvements

AndrewFerr:
Some type-checking improvements that make parts of the code more expressive & easier to work with.

Note that this is based on the config widget PR because it is the source of one of the type annotations added by this PR.

### DIFF
--- a/src/Widgets/BridgeWidgetApi.ts
+++ b/src/Widgets/BridgeWidgetApi.ts
@@ -156,7 +156,7 @@ export class BridgeWidgetApi {
         if (!connection.provisionerUpdateConfig || !connection.getProvisionerDetails)  {
             throw new ApiError("Connection type does not support updates", ErrCode.UnsupportedOperation);
         }
-        await connection.provisionerUpdateConfig(req.userId, req.body);
+        connection.provisionerUpdateConfig(req.userId, req.body);
         res.send(connection.getProvisionerDetails(true));
     }
 
--- a/changelog.d/505.misc
+++ b/changelog.d/505.misc
@@ -1,0 +1,1 @@
+Improve some type-checking in the codebase.
--- a/src/Connections/GithubRepo.ts
+++ b/src/Connections/GithubRepo.ts
@@ -1042,7 +1042,7 @@ ${event.release.body}`;
         }
     }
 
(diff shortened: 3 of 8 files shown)


src/Widgets/BridgeWidgetApi.ts
```diff
@@ -156,7 +156,7 @@ export class BridgeWidgetApi {
         if (!connection.provisionerUpdateConfig || !connection.getProvisionerDetails)  {
             throw new ApiError("Connection type does not support updates", ErrCode.UnsupportedOperation);
         }
-        await connection.provisionerUpdateConfig(req.userId, req.body);
+        connection.provisionerUpdateConfig(req.userId, req.body);
```
Half-Shot, line 159: Did this stop being promisey?

Half-Shot (approved):
One comment, otherwise LGTM

src/Widgets/BridgeWidgetApi.ts
AndrewFerr: Yes: https://github.com/matrix-org/matrix-hookshot/blob/main/src/Connections/IConnection.ts#L68

AndrewFerr:
Before submitting this, there's a type error added by this change that `yarn lint` missed which I'll try fixing.

https://github.com/matrix-org/matrix-hookshot/pull/505/files#diff-e9a8de0579bab9c5fea95a16abc38388d4f7b0712ea52c581238f6d4c6fccaa7R132

That change to the type of `filters` fixes lint errors, but causes mismatched-type errors when calling that function in places like this:
https://github.com/matrix-org/matrix-hookshot/blob/main/web/components/roomConfig/GitlabRepoConfig.tsx#L21

```
Argument of type 'GitLabTargetFilter' is not assignable to parameter of type 'Record<string, string> | Record<string, never> | undefined'.
  Type 'GitLabTargetFilter' is not assignable to type 'Record<string, never>'.
```

Basically, I want a type that can cover `{}`, `undefined`, and a `Record<string, string>`, but without using `{}` as a type, which the linter considers an error.

AndrewFerr:
Actually, I'll rebase this before merging, since it was based on the tip of the widgets branch.

PR #506, opened 2022-09-30
Move `web` tsconfig to let TS language server work

AndrewFerr:
This allows type-checking to work on Preact components used by *.tsx files in the `web` directory.  It works for me in VSCode, vim + YouCompleteMe, and Eclipse + Wild Web Developer.

This also includes an extra commit to fix type errors on imported images, since getting those errprs out of the way makes the most of the errors exposed by the tsconfig move.


Half-Shot (approved):
Thanks, this is much more sensible!

PR #507, opened 2022-10-03
Fix Jira widget request loop for logged-out users

AndrewFerr:
Without this change, viewing the Jira widget without being logged into Jira causes the widget to repeatedly make network requests.

Note that this intentionally lacks a changelog since it fixes an issue with a yet-to-be-released feature (the Jira widget).

### DIFF
--- a/changelog.d/507.bugfix
+++ b/changelog.d/507.bugfix
@@ -1,0 +1,1 @@
+Add a configuration widget for Jira.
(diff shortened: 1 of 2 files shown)


Half-Shot (approved):
Needs a changelog but otherwise good. If you copy paste the content of your changelog file for the JIRA feature PR, it will merge the two changelog entries into one upon release (to avoid bloating out the changelog file)

changelog.d/507.bugfix
```diff
@@ -0,0 +1 @@
+Add a configuration widget for Jira.
```
AndrewFerr, line 1: @Half-Shot This is the same [changelog entry](https://github.com/matrix-org/matrix-hookshot/pull/502/files#diff-76d69f49eb80d1d55e43a13b13cc35d753c6fc68569942622ee07b64b53e16d9R1) as #502's, but with a different filename; is that what you wanted?

Half-Shot: This is good.

PR #508, opened 2022-10-03
Reorganize the GitHub config widget

AndrewFerr:
- Group repos by org
- Paginate repo requests
- Remove redundant label next to chosen repo
- Remove redundant repo field below connected repo
- Include repos that user has doesn't have admin permissions for, to achieve parity with results from provisioning API

---

image
image

This allows the widget's search UI to match Dimension's:

image

For what the GH widget looked like before, [see here](https://github.com/matrix-org/matrix-hookshot/pull/420#issuecomment-1192876562).

### DIFF
--- a/web/components/roomConfig/GithubRepoConfig.tsx
+++ b/web/components/roomConfig/GithubRepoConfig.tsx
@@ -3,43 +3,69 @@ import { useState, useCallback, useEffect, useMemo } from "preact/hooks";
 import { BridgeAPI, BridgeConfig } from "../../BridgeAPI";
 import { ConnectionConfigurationProps, RoomConfig } from "./RoomConfig";
 import { ErrCode } from "../../../src/api";
-import { GitHubRepoConnectionState, GitHubRepoResponseItem, GitHubRepoConnectionTarget } from "../../../src/Connections/GithubRepo";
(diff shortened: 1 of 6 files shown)


web/components/roomConfig/GithubRepoConfig.tsx
```diff
@@ -52,29 +78,38 @@ const ConnectionSearch: FunctionComponent<{api: BridgeAPI, onPicked: (state: Git
...
-        () => results?.map(i => <option repo={i.state.repo} org={i.state.org} value={i.name} key={i.name} />),
+        () => results?.map(i => <option key={i.name} value={i.name} />),
         [results]
     );
 
 
     return <div>
         {isConnected === null && <p> Loading GitHub connection. </p>}
         {isConnected === false && <p> You are not logged into GitHub. </p>}
+        {isConnected === true && orgs?.length === 0 && <p> You do not have access to any GitHub organizations. </p>}
```
Half-Shot, line 104: Should we link to the docs on how to auth, or should that be left up to the admin/user to discover?

Half-Shot (approved):
One question, otherwise looks good.

web/components/roomConfig/GithubRepoConfig.tsx
AndrewFerr: May as well make the most of our docs.  However, I'll do it in a separate change to cover Jira & GitLab as well.

AndrewFerr: Tracked: https://github.com/matrix-org/matrix-hookshot/issues/510

Issue #510, opened 2022-10-03
Widgets should link to docs for adding auth when no connections are available


AndrewFerr:
If a user's GitLab/GitHub/Jira account has no access to any repos/projects, the user must tweak their account (or perhaps their Hookshot app installation) to get that access.  The widgets may as well provide a link to Hookshot's own docs to help guide that process.

PR #511, opened 2022-10-03
Preserve case-sensitivity of GitHub URLs

AndrewFerr:
Don't lower-case GitHub repos, orgs, usernames, etc when using them as room state event keys, room aliases, etc.

### DIFF
--- a/changelog.d/511.bugfix
+++ b/changelog.d/511.bugfix
@@ -1,0 +1,1 @@
+Preserve case-sensitivity of GitHub repository URLs.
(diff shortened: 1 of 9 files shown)


AndrewFerr:
Fixes #511 

This may de-lowercase more than is strictly necessary, but let's start here & peel away at what's unneeded (if anything).

Note that this change applies not only to room event state keys, but also to room aliases (which [the spec](https://github.com/matrix-org/matrix-hookshot/issues/510) doesn't call out as being case-insensitive).

AndrewFerr:
I'm afraid this still might break some places that expect lower-cased names; maybe it could be addressed by lower-casing only in safe places.  But I'll look into it after cutting the next release, as this isn't a gating issue.

Half-Shot (changes requested):
I think I may not have understood you earlier...the reason for lowercasing is to ensure we don't have clashes...this appears to be going in the opposite direction.

To the best of my knowledge, there isn't any reason to preserve case sensitivity in these urls:

https://github.com/MATRIX-ORG/MATRIX-HOOKSHOT/ is equivalent to https://github.com/matrix-org/matrix-hookshot and GitHub shouldn't care?

PR #515, opened 2022-10-03
Fix command prefix field in widgets

AndrewFerr:
This also fixes the comment body toggle for GitLab widgets.

### DIFF
--- a/src/Connections/GenericHook.ts
+++ b/src/Connections/GenericHook.ts
@@ -432,6 +432,7 @@ export class GenericHookConnection extends BaseConnection implements IConnection
                 hookId: this.hookId
             }
         );
+        this.state = validatedConfig;
     }
 
     public toString() {
--- a/src/Connections/GithubRepo.ts
+++ b/src/Connections/GithubRepo.ts
@@ -1133,6 +1133,7 @@ ${event.release.body}`;
     public async provisionerUpdateConfig(userId: string, config: Record<string, unknown>) {
         const validatedConfig = GitHubRepoConnection.validateState(config);
         await this.as.botClient.sendStateEvent(this.roomId, GitHubRepoConnection.CanonicalEventType, this.stateKey, validatedConfig);
+        this.state = validatedConfig;
     }
 
     public async onRemove() {
--- a/src/Widgets/BridgeWidgetApi.ts
+++ b/src/Widgets/BridgeWidgetApi.ts
@@ -156,7 +156,7 @@ export class BridgeWidgetApi {
         if (!connection.provisionerUpdateConfig || !connection.getProvisionerDetails)  {
             throw new ApiError("Connection type does not support updates", ErrCode.UnsupportedOperation);
         }
-        connection.provisionerUpdateConfig(req.userId, req.body);
+        await connection.provisionerUpdateConfig(req.userId, req.body);
         res.send(connection.getProvisionerDetails(true));
     }
 
--- a/changelog.d/515.bugfix
+++ b/changelog.d/515.bugfix
@@ -1,0 +1,1 @@
+Fix inactive "Command Prefix" field in configuration widgets.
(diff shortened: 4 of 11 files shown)


src/Connections/GenericHook.ts
```diff
@@ -432,6 +432,7 @@ export class GenericHookConnection extends BaseConnection implements IConnection
                 hookId: this.hookId
             }
         );
+        this.state = validatedConfig;
```
Half-Shot, line 435: what, how, what, how did this work

src/Connections/GithubRepo.ts
```diff
@@ -1133,6 +1133,7 @@ ${event.release.body}`;
     public async provisionerUpdateConfig(userId: string, config: Record<string, unknown>) {
         const validatedConfig = GitHubRepoConnection.validateState(config);
         await this.as.botClient.sendStateEvent(this.roomId, GitHubRepoConnection.CanonicalEventType, this.stateKey, validatedConfig);
+        this.state = validatedConfig;
```
Half-Shot, line 1136: Oh, I think because we sent a state event to the HS, it sends a event back to us and updates the connection for us...I don't know whether this is needed.

Half-Shot (changes requested):
Refs look good. I suspect the provisioner update functions work as intended...

src/Widgets/BridgeWidgetApi.ts
```diff
@@ -156,7 +156,7 @@ export class BridgeWidgetApi {
         if (!connection.provisionerUpdateConfig || !connection.getProvisionerDetails)  {
             throw new ApiError("Connection type does not support updates", ErrCode.UnsupportedOperation);
         }
-        connection.provisionerUpdateConfig(req.userId, req.body);
+        await connection.provisionerUpdateConfig(req.userId, req.body);
```
AndrewFerr, line 159: @Half-Shot You were correct to have [questioned](https://github.com/matrix-org/matrix-hookshot/pull/505/commits/08c70038d1ce14af1701b8ac01bec0b2823b5c25#diff-2aa3eec27e7c4d9bb2c3598bdcf50971052c917cb826c58bb00884430d1d3aa8) the promise-ness of this :slightly_smiling_face:   Turns out the real culprit was an improper return type on the method declaration (`void` instead of `Promise<void>`).

src/Connections/GithubRepo.ts
AndrewFerr: It is, because after a save, the widget UI updates to match the state object returned by the "update" widget API call.  Without updating the state object here, the widget API response returns the state object from before the update, and the widget UI won't show the change you just made.

src/Connections/GenericHook.ts
AndrewFerr: Should it not...? `state` and `validatedConfig` are each a `GenericHookConnectionState`.

src/Connections/GithubRepo.ts
AndrewFerr: Actually, the stale updates are partially fixed by correctly awaiting on the update, but setting the state here still prevents a flicker in the UI:

[hookshotWidgetPrefixFlicker.webm](https://user-images.githubusercontent.com/3271094/193683243-5609971a-ce81-46d8-aedc-8b2f38f0acdd.webm)

AndrewFerr: Turns out there's more to this after all -- will return with a full fix

AndrewFerr: ecb3796 is a quick fix for this:

[hookshotWidgetPrefixNoFlicker.webm](https://user-images.githubusercontent.com/3271094/193728786-147f1e1c-07de-4875-a945-79c84cc5f091.webm)

Updating `this.state` in each of the `provisionerUpdateConfig`s is still useful, as the current flow of saving a config & the widget picking it up is quite racy:
1. Widget makes widget API request to update the connection's config state
2. Bridge calls `provisionerUpdateConfig` to send a state event with the new config state, and responds with 200 + the config state object
3. Bridge eventually sees that config state again as a state event change from the homeserver, and updates the connection's config state JS object
4. Widget makes new widget API request to get the latest config states of all connections

Steps 3 and 4 each happen asynchronously and can occur in any order, meaning that if the config state isn't eagerly saved in Step 2, the bridge may answer the widget API request in Step 4 with a stale state before the HS sends the state event back to the bridge in Step 3.

Making matters more confusing is that the bridge's response in Step 2 includes the config state object, which 1) is a stale object if it isn't eagerly saved, and 2) isn't used by the widget's response logic anyways.

With all that said, the current change should be plenty for avoiding these problems.  I've opened #516 to track tightening up the handling of bot-sent state events, when there's time/interest to look into it.

Half-Shot (approved):

PR #523, opened 2022-10-07
Improvements to GitHub auth

Half-Shot:
While trying to investigate why the bridge stores an empty token when trying to perform a token refresh, I fixed these issues. For whatever reason, I cannot repro this problem on my local setup.

Having tested this on element.io, it doess seem to fix the issue.


AndrewFerr (approved):
Looks good, just have some nitpicks

Issue #532, opened 2022-10-11
Add Release Events to Jira Bridge


gaelgatelement:
It could be great to also support Release Events, not only Issue creation events.

AndrewFerr:
Fixed by #534.

PR #534, opened 2022-10-14
Support Jira version events (WIP)

AndrewFerr:
The closest thing Jira has to release events is "version" events, which encompass created, updated, and released versions.
https://docs.atlassian.com/software/jira/docs/api/REST/9.3.0/#api/2/version

The biggest missing piece is a way to link a version's project to a project bound to a room: rooms save projects by their URL, while version event webhook payloads only have their ID (which hookshot used to store but no longer does).


AndrewFerr:
This addresses #532 btw

AndrewFerr:
This also:
- fixes "issue created" events, which were getting ignored due to code expecting the wrong webhook type string (`issue.created` instead of `issue_created`)
- unlocks support for "issue updated" events, which was implemented but not used

Screenshot from 2022-10-18 01-00-41

Screenshot from 2022-10-18 01-00-25

AndrewFerr:
Lastest push just rebases onto the tip of `main` to get the CI fixes.

AndrewFerr:
This PR is now split in two: issue event fixes are in #543, and version event support is here & is applied on top of the issue event fixes.

The content of this PR is the same as it was before that reorganization (aside from some trivial changes & being updated onto the latest tip of `main`).

AndrewFerr:
Final push is just a rebase to avoid a fake merge conflict.  It produces an empty diff with the pre-rebased version.

PR #535, opened 2022-10-14
Fix Docker builds hanging/OOMing

AndrewFerr:


Half-Shot (approved):
If it works, woo!

Half-Shot:
This looks like `cargo` taking too long to download crates, perhaps we need to fix that timeout too.

PR #536, opened 2022-10-17
Skip GitLab MR reply comments

Half-Shot:
GitLab uses `discussion_id` to chain together replies. We can't really tell what position we are *in* the thread, so we just have to assume that if we've not seen the discussion before that we are the first note, and subsequent notes are child notes.

This also changes the code so that we skip sending a notice into the room if the user has disabled comment-only notices.

### DIFF
--- a/src/Connections/GitlabRepo.ts
+++ b/src/Connections/GitlabRepo.ts
@@ -655,7 +655,7 @@ ${data.description}`;
             if (commentNotes) {
                 existing.commentNotes = [...(existing.commentNotes ?? []), ...commentNotes];
             }
-            existing.commentCount = existing.commentCount + opts.commentCount;
+            existing.commentCount += opts.commentCount;
             if (!opts.skip) {
                 existing.skip = false;
             }
--- a/src/Connections/GitlabRepo.ts
+++ b/src/Connections/GitlabRepo.ts
@@ -286,6 +286,7 @@ export class GitLabRepoConnection extends CommandConnection<GitLabRepoConnection
         author: string,
         timeout: NodeJS.Timeout,
         approved?: boolean,
+        skip?: boolean,
     }>();
 
     /**
@@ -639,6 +640,10 @@ ${data.description}`;
             commentCount: number,
             commentNotes?: string[],
             approved?: boolean,
+            /**
+             * If the MR contains only comments, skip it.
+             */
+            skip: boolean,
         }
     ) {
         const { commentCount, commentNotes, approved } = opts;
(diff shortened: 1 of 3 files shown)


src/Connections/GitlabRepo.ts
```diff
@@ -285,8 +286,16 @@ export class GitLabRepoConnection extends CommandConnection<GitLabRepoConnection
         author: string,
         timeout: NodeJS.Timeout,
         approved?: boolean,
+        skip?: boolean,
     }>();
 
+    /**
+     * GitLab provides NO threading information in it's webhook response objects,
```
jaller94, line 293:
```suggestion
     * GitLab provides NO threading information in its webhook response objects,
```

src/Connections/GitlabRepo.ts
```diff
@@ -671,23 +688,42 @@ ${data.description}`;
...
+                skip: false,
             }
         );
     }
 
+    private shouldHandleMRComment(event: IGitLabWebhookNoteEvent) {
+        // Check to see if this line has had a comment before
+        if (event.object_attributes.discussion_id) {
+            if (this.mergeRequestSeenDiscussionIds.has(event.object_attributes.discussion_id)) {
+                // If it has, this is probably a reply. Replies are noise, skip em.
```
jaller94, line 701:
```suggestion
                // If it has, this is probably a reply. Skip repeated replies.
```

src/Connections/GitlabRepo.ts
```diff
@@ -642,13 +655,17 @@ ${data.description}`;
             if (commentNotes) {
                 existing.commentNotes = [...(existing.commentNotes ?? []), ...commentNotes];
             }
-            existing.commentCount = opts.commentCount;
+            existing.commentCount = existing.commentCount + opts.commentCount;
```
jaller94, line 658: This suggestion is just fancy syntax sugar. No need to apply.
```suggestion
            existing.commentCount += opts.commentCount;
```

jaller94 (approved):
Looks good to me.

I'm a bit surprised that Hookshot depends on `@alloc/quick-lru` instead of the upstream `quick-lru` (which became an ESM in version 6).

Half-Shot:
>I'm a bit surprised that Hookshot depends on @alloc/quick-lru instead of the upstream quick-lru (which became an ESM in version 6).

We can't use ESM, as it introduces a bunch of annoying incompatibilities. Instead of trying to refactor the entire project to be ESM compatible, we've settled for just using a cJS library for the moment.

PR #539, opened 2022-10-17
Check GitLab Repo access level via a better API

### LINKED ISSUE #538: Improve Gitlab permission checking
Given:
* I am a member of a group `o` on a Gitlab instnace
* Group `o` is given owner permissions on another group `i`
* Group `i` has a sub-project `s`
* `/api/v4/projects/<project id>` with my Gitlab PAT returns:
```
  "permissions": {
    "project_access": null,
    "group_access": null
  }
```

This means that Hookshot decides I don't have the access level to setup the webhooks on project `s`

Options:
* Assume Gitlab 15.4 and consider using https://docs.gitlab.com/ee/api/members.html#get-a-member-of-a-group-or-project-including-inherited-and-invited-members which should make this visible via https://gitlab.com/gitlab-org/gitlab/-/merge_requests/83214
* If `/api/v4/projects/<project id>` returns something (i.e. I can view the project) then attempt to add the hook regardless and interpret the response rather than deciding I don't have access pre-emptively

Half-Shot:
Fixes #538 

Use a better API for checking access levels, that accounts for parent group permissions rather than just project permissions.

### DIFF
--- a/src/Gitlab/Client.ts
+++ b/src/Gitlab/Client.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosError } from "axios";
+import axios from "axios";
 import { GitLabInstance } from "../Config/Config";
 import { GetIssueResponse, GetUserResponse, CreateIssueOpts, CreateIssueResponse, GetIssueOpts, EditIssueOpts, GetTodosResponse, EventsOpts, CreateIssueNoteOpts, CreateIssueNoteResponse, GetProjectResponse, ProjectHook, ProjectHookOpts, AccessLevel, SimpleProject } from "./Types";
 import { Logger } from "matrix-appservice-bridge";
--- a/src/Gitlab/Client.ts
+++ b/src/Gitlab/Client.ts
@@ -1,4 +1,4 @@
-import axios from "axios";
+import axios, { AxiosError } from "axios";
 import { GitLabInstance } from "../Config/Config";
 import { GetIssueResponse, GetUserResponse, CreateIssueOpts, CreateIssueResponse, GetIssueOpts, EditIssueOpts, GetTodosResponse, EventsOpts, CreateIssueNoteOpts, CreateIssueNoteResponse, GetProjectResponse, ProjectHook, ProjectHookOpts, AccessLevel, SimpleProject } from "./Types";
 import { Logger } from "matrix-appservice-bridge";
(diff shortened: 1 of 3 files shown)


src/Gitlab/Client.ts
```diff
@@ -134,6 +134,17 @@ export class GitLabClient {
         }
     }
 
+    public async getProjectAccessLevel(id: ProjectId): Promise<AccessLevel> {
+        try {
+            const me = await this.user();
+            const { data } = await axios.get(`api/v4/projects/${encodeURIComponent(id)}/members/${me.id}`, this.defaultConfig);
+            return data.access_level as AccessLevel;
+        } catch (ex) {
+            log.warn(`Failed to get project access level:`, ex);
```
jaller94, line 143: Do axios errors contain the HTTP headers (auth token secrets)?

src/Gitlab/Client.ts
```diff
@@ -134,6 +134,17 @@ export class GitLabClient {
         }
     }
 
+    public async getProjectAccessLevel(id: ProjectId): Promise<AccessLevel> {
+        try {
+            const me = await this.user();
+            const { data } = await axios.get(`api/v4/projects/${encodeURIComponent(id)}/members/${me.id}`, this.defaultConfig);
+            return data.access_level as AccessLevel;
```
jaller94, line 141: I wish there was some type and value validation of access_level.
That way it would be easier to spot if the API breaks.

src/Gitlab/Client.ts
```diff
@@ -134,6 +134,17 @@ export class GitLabClient {
         }
     }
 
+    public async getProjectAccessLevel(id: ProjectId): Promise<AccessLevel> {
+        try {
+            const me = await this.user();
```
jaller94, line 139: I wish there was some type and value validation in `user()`.
This flow just assumes that me is a valid object and has a valid attribute `id`.

Given it's all wrapped in a `catch (err) {console.warn(err)}`, it might be ok for debugging, in case the API breaks...

jaller94 (approved):
I wish there was more type validation for API responses, but it might be ok... It looks like an invalid response will just fail on the next step and likely give a reasonable-ish error message.

jaller94 (changes requested):
Just saw the TypeScript errors (and newsfile missing). Those need fixing before the final review.

src/Gitlab/Client.ts
Half-Shot: Added type checking, though we probably need to figure out how to do this as a whole.

jaller94 (approved):
Looks good to me.
Too bad the logging of secrets through axios errors cannot be easily fixed without missing out on valueable error messages.

Half-Shot:
Ah bother, I meant to merge this into the last release.

src/Gitlab/Client.ts
Half-Shot: I think it's okay for now, though point made.

Issue #421, opened 2022-07-22
Command prefixes can clash


AndrewFerr:
**Describe the bug**
When there are multiple connections in the same room, it is possible for them to share the same command prefix (`!gl`, `!gh`, etc).  If a command with a shared prefix is run, it will only be processed against one of the connections.  This prevents being able to run commands for some of the connections that share a command prefix.

**To Reproduce**
1. Connect two GitLab projects to a room, and do not customize the command prefix from its default of `!gl`
2. Try to run `!gl` commands in that room
3. Notice that only one of the projects accepts commands

**Expected behavior**
It should be forbidden for multiple connections to share the same command prefix.  If adding a new connection with a default prefix clashes with one already in use within a room, either the prefix should be modified to be unique, or the action of adding a connection should require the user to set a custom, unique prefix for it.

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

HarHarLinks:
see also #244: currently command prefix are confusing anyway

Half-Shot:
The workaround at the moment is adding your own prefix, though I would like to see the ConnectionManager basically handle this automatically. I like the idea of the connection manager handling prefixes where:

Connections with matching prefixes automatically gain a identifier prefix like `!gl` -> `!gl my-org/my-project` and using `!gl` will just warn you to use the right one. It would still be nice to set a default though, as in lots of places we tend to expect one. This could be the case of adding some state to the connection to denote that it is the default one.

AndrewFerr:
Even after #540, another issue remains: adding connections via bot commands (instead of the widget or provisioning API) allows creating connections with a prefix that's already in use.

Part of the problem is that bot commands don't allow setting a custom prefix for a new connection.

PR #540, opened 2022-10-18
Allow multiple Jira connections; stop prefix clash

### LINKED ISSUE #533: Support bridging multiple jira projects in the same room
At the moment, running the following commands multiple times in the same room  : 

`!hookshot jira project <project>`

Results in : 

```
WARN 08:09:44:493 [CommandConnection] Failed to handle command: Error: Cannot support multiple connections of the same type yet
    at Function.provisionConnection (/bin/matrix-hookshot/Connections/JiraProject.js:74:19)
    at SetupConnection.onJiraProject (/bin/matrix-hookshot/Connections/SetupConnection.js:125:52)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async handleCommand (/bin/matrix-hookshot/BotCommands.js:105:32)
    at async SetupConnection.onMessageEvent (/bin/matrix-hookshot/Connections/CommandConnection.js:41:31)
    at async Bridge.onRoomMessage (/bin/matrix-hookshot/Bridge.js:606:21)
```

AndrewFerr:
Fixes #533
Mitigates #421

----
image

Attempting to add a connection with a command prefix that's already in use for the room will fail:
Screenshot from 2022-10-18 02-22-00

### DIFF
--- a/src/ConnectionManager.ts
+++ b/src/ConnectionManager.ts
@@ -252,8 +252,11 @@ export class ConnectionManager extends EventEmitter {
     public validateCommandPrefix(roomId: string, config: IConnectionState, currentConnection?: IConnection) {
         if (config.commandPrefix === undefined) return;
         for (const c of this.getAllConnectionsForRoom(roomId)) {
-            if (c != currentConnection && c.conflictsWithCommandPrefix(config.commandPrefix)) {
-                throw new ApiError(`Command prefix "${config.commandPrefix}" is already used in this room. Please choose another prefix.`);
+            if (c != currentConnection && c.conflictsWithCommandPrefix?.(config.commandPrefix)) {
+                throw new ApiError(`Command prefix "${config.commandPrefix}" is already used in this room. Please choose another prefix.`, ErrCode.ConflictingConnection, -1, {
+                        existingConnection: c.getProvisionerDetails?.(),
+                    }
+                );
             }
         }
     }
--- a/src/ConnectionManager.ts
+++ b/src/ConnectionManager.ts
@@ -7,7 +7,7 @@
 import { Appservice, StateEvent } from "matrix-bot-sdk";
 import { CommentProcessor } from "./CommentProcessor";
 import { BridgeConfig, BridgePermissionLevel, GitLabInstance } from "./Config/Config";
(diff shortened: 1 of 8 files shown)


src/Connections/CommandConnection.ts
```diff
@@ -31,6 +31,10 @@ export abstract class CommandConnection<StateType extends IConnectionState = ICo
         return (this.state.commandPrefix || this.defaultCommandPrefix) + " ";
     }
 
+    public conflictsWithCommandPrefix(commandPrefix: string) {
+        return this.commandPrefix == commandPrefix + " ";
```
Half-Shot, line 35: Should we `toLower()` both of these to check for casing clashes?

src/Connections/BaseConnection.ts
```diff
@@ -18,4 +18,9 @@ export abstract class BaseConnection {
     public get priority(): number {
         return -1;
     }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    public conflictsWithCommandPrefix(commandPrefix: string) {
```
Half-Shot, line 23: Thought: If this is effectively optional, why not make the function implementation optional and just call it with `conflictsWithCommandPrefix?.()`. This would allow the caller to decide what to do with a missing implementation, and avoids the eslint hacks.

src/ConnectionManager.ts
```diff
@@ -249,6 +249,15 @@ export class ConnectionManager extends EventEmitter {
         return this.connections.find((c) => c.connectionId === connectionId && c.roomId === roomId);
     }
 
+    public validateCommandPrefix(roomId: string, config: IConnectionState, currentConnection?: IConnection) {
+        if (config.commandPrefix === undefined) return;
+        for (const c of this.getAllConnectionsForRoom(roomId)) {
+            if (c != currentConnection && c.conflictsWithCommandPrefix(config.commandPrefix)) {
+                throw new ApiError(`Command prefix "${config.commandPrefix}" is already used in this room. Please choose another prefix.`);
```
Half-Shot, line 256: Should this ApiError return something like a conflict error rather than the default unknown error code?

Half-Shot (approved):
Looks good, but some thoughts attached.

src/Connections/CommandConnection.ts
AndrewFerr: IMO no, because command prefixes themselves are case-sensitive.

src/Connections/BaseConnection.ts
AndrewFerr: Good idea, will do

src/ConnectionManager.ts
AndrewFerr: Yes it should; I had missed the status code property of ApiError.

Half-Shot (approved):

PR #543, opened 2022-10-21
Fix & refactor Jira issue event handling

AndrewFerr:
- Refactor Jira event handling to be more like GitHub & GitLab
- Fix silently-ignored Jira events due to expecting wrong type strings
- Update UI for Jira events

---

This is the portion of #534 unrelated to version events, split into its own PR for better visibility.


Half-Shot (approved):
Looks good, thanks!

PR #546, opened 2022-10-26
Maintainer Gitlab perms instead of Developer.

### LINKED ISSUE #545: Improve hookshook error message when gitlab token doesn't have the "maintainer" role e.g "Failed to create connection because of insufficient permissions"
* The gitlab token needs the gitlab `maintainer` role ; the gitlab `developer` role is not sufficient.
image

rtanglao:
* Gnome needed to use Gitlab `maintainer` permissions, partially fixes #545

### DIFF
--- a/docs/usage/room_configuration/gitlab_project.md
+++ b/docs/usage/room_configuration/gitlab_project.md
@@ -13,7 +13,7 @@ To set up a connection to a GitLab project in a new room:
 2. Invite the bridge bot (e.g. `@hookshot:example.com`).
 3. Give the bridge bot moderator permissions or higher (power level 50) (or otherwise configure the room so the bot can edit room state).
 4. Send the command `!hookshot gitlab project https://mydomain/my/project`.
-5. If you have permission to bridge this repo, the bridge will respond with a confirmation message. (Users with `Developer` permissions or greater can bridge projects.)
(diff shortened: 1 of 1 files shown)


AndrewFerr (changes requested):
There is no "Moderator" role, but there is a "Maintainer" one.

I also want to confirm that it's indeed forbidden for "Developer"s to simply _list_ available webhooks, which should be sufficient for preparing a Hookshot connection for a repo's existing webhooks.

If there isn't, we'll also have to update the bot's error message of "You must at least have developer access to bridge this project".

rtanglao:
> There is no "Moderator" role, but there is a "Maintainer" one.
Oops fixed it!

> I also want to confirm that it's indeed forbidden for "Developer"s to simply _list_ available webhooks, which should be sufficient for preparing a Hookshot connection for a repo's existing webhooks.
> 
didn't touch this, how do I confirm this?

> If there isn't, we'll also have to update the bot's error message of "You must at least have developer access to bridge this project
i'll leave this to you @AndrewFerr :-)

rtanglao:
closed by accident, please re-review :-)

AndrewFerr:
As it turns out, the Maintainer role is only needed for Hookshot to add a webhook to the project.  Given that disabling automatic webhook creation in the config is possible & would bypass this error, we should allow this missing permission & just behave as though webhook creation was disabled.

That and more is handled in https://github.com/matrix-org/matrix-hookshot/pull/567.

PR #553, opened 2022-11-01
Don't lower-case GitLab URLs with bot command 

AndrewFerr:
This is a simpler fix than #511 because, although GitLab project paths seem to be case-sensitive, several scenarios effectively ignore their casing anyways [1], [2], so it's reasonable for Hookshot do to the same.

[1] https://gitlab.com/gitlab-org/gitlab/-/issues/14533
[2] https://gitlab.com/gitlab-org/gitlab-foss/-/issues/30997

### DIFF
--- a/src/ConnectionManager.ts
+++ b/src/ConnectionManager.ts
@@ -205,7 +205,7 @@ export class ConnectionManager extends EventEmitter {
 
     public getConnectionsForGitLabRepo(pathWithNamespace: string): GitLabRepoConnection[] {
         pathWithNamespace = pathWithNamespace.toLowerCase();
-        return this.connections.filter((c) => (c instanceof GitLabRepoConnection && c.path === pathWithNamespace)) as GitLabRepoConnection[];
+        return this.connections.filter((c) => (c instanceof GitLabRepoConnection && c.path.toLowerCase() === pathWithNamespace)) as GitLabRepoConnection[];
     }
 
     public getConnectionsForJiraProject(project: JiraProject): JiraProjectConnection[] {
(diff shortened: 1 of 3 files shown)


src/ConnectionManager.ts
```diff
@@ -205,7 +205,7 @@ export class ConnectionManager extends EventEmitter {
 
     public getConnectionsForGitLabRepo(pathWithNamespace: string): GitLabRepoConnection[] {
         pathWithNamespace = pathWithNamespace.toLowerCase();
-        return this.connections.filter((c) => (c instanceof GitLabRepoConnection && c.path === pathWithNamespace)) as GitLabRepoConnection[];
+        return this.connections.filter((c) => (c instanceof GitLabRepoConnection && c.path.toLowerCase() === pathWithNamespace)) as GitLabRepoConnection[];
```
Half-Shot, line 208: I'd make `c.path` itself lowercase in `GitLabRepoConnection`, and check all usages of it. We shouldn't be having mixed-case state keys as a rule.

AndrewFerr: That would cause a force-lowercased path to be sent to `GitLabClient.issues.create`.  Could that be a problem?

Half-Shot: > That would cause a force-lowercased path to be sent to `GitLabClient.issues.create`. Could that be a problem?

Should be fine :+1:

Half-Shot (approved):

src/ConnectionManager.ts
AndrewFerr: Tested & confirmed that `GitLabClient.issues.create` is case insensitive after all :+1:

AndrewFerr:
I'll merge this now to fix the problem with GitLab webhooks being ignored, but there will be a follow-up change to prevent adding multiple connections for the same GitLab project by changing its casing (i.e. you can add connections for `path/repo`, `Path/Repo`, `pAtH/RePo`, and they'll all respond to events on the same project, whatever its path's actual casing is).

PR #554, opened 2022-11-01
GitHub assign quality of life.

Half-Shot:
Just to make the assign command nicer.

### DIFF
--- a/src/Connections/GithubRepo.ts
+++ b/src/Connections/GithubRepo.ts
@@ -649,8 +649,8 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
         }
     }
 
-    @botCommand("assign", "Assign an issue to a user", ["number", "...users"], [], true)
-    public async onAssign(userId: string, number: string, ...users: string[]) {
+    @botCommand("assign", "Assign an issue to a user. If number is ommitted, the latest issue is used. If users is omitted, you are assigned.", [], ["number", "...users"], true)
+    public async onAssign(userId: string, number?: string, ...users: string[]) {
         const octokit = await this.tokenStore.getOctokitForUser(userId);
         if (!octokit) {
             throw new NotLoggedInError();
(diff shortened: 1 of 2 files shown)


src/Connections/GithubRepo.ts
```diff
@@ -649,8 +649,8 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
         }
     }
 
-    @botCommand("assign", "Assign an issue to a user", ["number", "...users"], [], true)
-    public async onAssign(userId: string, number: string, ...users: string[]) {
+    @botCommand("assign", "Assign an issue to a user. If number is ommitted, the latest issue is used. If users is omitted, you are assigned.", [], ["number", "...users"], true)
```
AndrewFerr, line 652: How about backticking the parameter names to make them jump out better?
```suggestion
    @botCommand("assign", "Assign an issue to a user. If `number` is ommitted, the latest issue is used. If `users` is omitted, you are assigned.", [], ["number", "...users"], true)
```
I've confirmed that it works:
image

src/Connections/GithubRepo.ts
```diff
@@ -660,10 +660,30 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
             users = users[0].split(",");
         }
 
+        if (users.length === 0) {
+            // Assume self.
+            users = [(await octokit.users.getAuthenticated()).data.login];
+        }
+
+        let issueNumber = number && parseInt(number, 10);
+        if (!issueNumber) {
```
AndrewFerr, line 669: Does it matter that passing 0 or a non-number value is treated the same as passing no number?

If it's worth anything, passing a negative number explicitly fails.

AndrewFerr (approved):
Optional request: it would be nice if the bot would echo which issue/assignee was used.

src/Connections/GithubRepo.ts
```diff
@@ -678,7 +678,9 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
                 throw new CommandError('No issues found', 'There are no issues on this repository');
             }
             issueNumber = topIssue.number;
-        }
+        } else {
+            issueNumber = parseInt(number, 10);
```
AndrewFerr, line 682: If `number` is a non-numberable string, this will set `issue_number` to `NaN`. Is that ok?

Half-Shot: Yeah, we can let GitHub explode as we did before.

PR #556, opened 2022-11-01
Display true casing of GitLab paths

AndrewFerr:

### DIFF
--- a/changelog.d/556.feature
+++ b/changelog.d/556.feature
@@ -1,0 +1,1 @@
+Display GitLab project paths with their true letter casing.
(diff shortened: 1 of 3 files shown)


changelog.d/556.feature
```diff
@@ -0,0 +1 @@
+Display GitLab project paths with their true letter casing.
```
Half-Shot, line 1: Display where? And this might be a `misc` rather than a `feature`?

Half-Shot (approved):

changelog.d/556.feature
AndrewFerr: In the widget's listing of connected projects & the bot's response to `gitlab project` (i.e. "Room configured to bridge $path").

AndrewFerr: Also, I didn't mark this as `misc` because it's user-facing.  Is `misc` also for minor touchups, even if they are visible?

AndrewFerr: Sticking with `feature` for now

PR #558, opened 2022-11-01
Don't provision command-created connections twice

AndrewFerr:
Without this change, adding a connection with a bot command creates the connection, but doesn't store it in memory (by pushing it into the connection manager's list of known connections).  Thus, what would actually store the connection in memory was the bot responding to its own room state event, which made it provision the connection again from scratch.


AndrewFerr:
The force push is just a rebase on the latest `main`.

Issue #559, opened 2022-11-02
Generic Webhook fails with `Bad JSON value: float`


benbz:
I've got a process that passes a JSON blob (including a `text` field) through to Hookshot as JSON. Sometimes these fail (silently from the caller's PoV) with the following in the Hookshot logs

```
INFO 00:08:05:130 [GenericHookConnection] onGenericHook <roomid> dddcd73f-4568-439a-9255-49edbefc30a1
ERROR 00:08:05:178 [MatrixHttpClient] (REQ-39033) { errcode: 'M_BAD_JSON', error: 'Bad JSON value: float' }
WARN 00:08:05:179 [Bridge] Failed to handle generic webhook Error: Failed to send Matrix message
    at MessageSenderClient.sendMatrixMessage (/bin/matrix-hookshot/MatrixSender.js:89:15)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async GenericHookConnection.onGenericHook (/bin/matrix-hookshot/Connections/GenericHook.js:308:9)
    at async /bin/matrix-hookshot/Bridge.js:387:25
    at async Promise.all (index 0)
    at async LocalMQ.<anonymous> (/bin/matrix-hookshot/Bridge.js:368:13)
```

The JSON blob does have some floats in it but I'd have hoped https://github.com/matrix-org/matrix-hookshot/blob/main/src/Connections/GenericHook.ts#L68-L99 was filtering those out.

I can provide the full JSON blob on request.

PR #560, opened 2022-11-02
Increase SANITIZE_MAX_BREADTH/SANITIZE_MAX_DEPTH for GenericHook

Half-Shot:
I suspect #559 is failing because the body falls just outside our safe range. The safe ranges were chosen arbitrarily, but we can probably increase them without a huge perf penalty.

### DIFF
--- a/src/Connections/GenericHook.ts
+++ b/src/Connections/GenericHook.ts
@@ -73,27 +73,27 @@ export class GenericHookConnection extends BaseConnection implements IConnection
      * If the object contains more than `SANITIZE_MAX_BREADTH` entries, the remaining entries will not be checked.
      * 
      * @param data The data to santise
-     * @param depth The depth of the current object relative to the root.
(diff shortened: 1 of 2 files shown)


src/Connections/GenericHook.ts
```diff
@@ -73,28 +73,35 @@ export class GenericHookConnection extends BaseConnection implements IConnection
...
+
+        let objBreadth = 0;
         const obj: Record<string, unknown> = { ...data };
         for (const [key, value] of Object.entries(data)) {
-            breadth++;
-            if (breadth > SANITIZE_MAX_BREADTH) {
-                break;
-            }
-            obj[key] = this.sanitiseObjectForMatrixJSON(value, depth + 1);
+            obj[key] = this.sanitiseObjectForMatrixJSON(value, depth + 1, ++objBreadth);
```
AndrewFerr, line 102: Ditto

AndrewFerr (approved):

PR #561, opened 2022-11-02
Prevent users without permissions from inviting the bot

Half-Shot:

### DIFF
--- a/changelog.d/561.bugfix
+++ b/changelog.d/561.bugfix
@@ -1,1 +1,1 @@
-The bot no longer accepts invites from user who do not have permission to use it.
+The bot no longer accepts invites from users who do not have permission to use it.
--- a/changelog.d/561.bugfix
+++ b/changelog.d/561.bugfix
@@ -1,0 +1,1 @@
+The bot no longer accepts invites from user who do not have permission to use it.
--- a/src/Bridge.ts
+++ b/src/Bridge.ts
@@ -783,7 +783,7 @@ export class Bridge {
         }
 
         // Don't accept invites from people who can't do anything
-        if (this.config.checkPermissionAny(event.sender, BridgePermissionLevel.login)) {
+        if (!this.config.checkPermissionAny(event.sender, BridgePermissionLevel.login)) {
             return this.as.botIntent.underlyingClient.kickUser(this.as.botUserId, roomId, "You do not have permission to inite this bot");
         }
 
--- a/src/Bridge.ts
+++ b/src/Bridge.ts
@@ -779,7 +779,7 @@ export class Bridge {
         log.info(`Got invite roomId=${roomId} from=${event.sender} to=${event.state_key}`);
         // Room joins can fail over federation
         if (event.state_key !== this.as.botUserId) {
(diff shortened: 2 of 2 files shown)


changelog.d/561.bugfix
```diff
@@ -0,0 +1 @@
+The bot no longer accepts invites from user who do not have permission to use it.
```
AndrewFerr, line 1:
```suggestion
The bot no longer accepts invites from users who do not have permission to use it.
```

src/Bridge.ts
```diff
@@ -781,6 +781,12 @@ export class Bridge {
         if (event.state_key !== this.as.botUserId) {
             return this.as.botIntent.underlyingClient.kickUser(this.as.botUserId, roomId, "Bridge does not support DMing ghosts");
         }
+
+        // Don't accept invites from people who can't do anything
+        if (this.config.checkPermissionAny(event.sender, BridgePermissionLevel.login)) {
+            return this.as.botIntent.underlyingClient.kickUser(this.as.botUserId, roomId, "You do not have permission to inite this bot");
```
AndrewFerr, line 787:
```suggestion
            return this.as.botIntent.underlyingClient.kickUser(this.as.botUserId, roomId, "You do not have permission to invite this bot.");
```

src/Bridge.ts
```diff
@@ -781,6 +781,12 @@ export class Bridge {
         if (event.state_key !== this.as.botUserId) {
             return this.as.botIntent.underlyingClient.kickUser(this.as.botUserId, roomId, "Bridge does not support DMing ghosts");
         }
+
+        // Don't accept invites from people who can't do anything
+        if (this.config.checkPermissionAny(event.sender, BridgePermissionLevel.login)) {
```
AndrewFerr, line 786:
```suggestion
        if (!this.config.checkPermissionAny(event.sender, BridgePermissionLevel.login)) {
```

src/Bridge.ts
AndrewFerr, line 787: FWIW, some clients (we all know which one) don't render the content of room leaves, so consider having the bot accept the join, post a goodbye message, and then leave the room.  (If this is done, there may have to be something to cleanly handle all of the bot's `"room.join"` callbacks.)

If it's not important to display a message, this can instead just use `leaveRoom`.

Also, may as well convert `botIntent.underlyingClient` to `botClient` for better readability.

AndrewFerr (changes requested):

src/Bridge.ts
AndrewFerr: Actually, `botIntent.underlyingClient` appears all throughout the code, so maybe there can be a dedicated PR to change all of them at once.

Half-Shot: >don't render the content of room leaves, so consider having the bot accept the join, post a goodbye message, and then leave the room

I'd consider that a client that doesn't conform to the spec. Joins are **not** free (imagine being a small homeserver getting an invite to HQ), so the reason for a leave is to avoid that expensive operation.

AndrewFerr (approved):

PR #563, opened 2022-11-02
Remove trailing slashes from GitLab instance URLs

### LINKED ISSUE #562: Can't use bot command to add GitLab projects on instances with a trailing slash
To reproduce:
- Edit your Hookshot config to have a GitLab instance with a trailing slash, such as:
```
gitlab:
    instances:
        myinstance:
            url: https://gitlab.com/
```
- Set up Hookshot with this instance in the usual way (give it your `personaltoken` in a DM, invite it to a room, etc)
- Try adding a connection for a project on that instance via `!hookshot gitlab project myinstance <url>`

This will fail with
> Failed to handle command: Could not determine if the user has access to this project, does the project exist?

Internally, it happens because some code strips away one character too many from the project path before calling GitLab API endpoints, all because of that pesky trailing slash.

AndrewFerr:
Fixes #562

### DIFF
--- a/src/Config/Config.ts
+++ b/src/Config/Config.ts
@@ -209,6 +209,13 @@ export class BridgeConfigGitLab {
         this.instances = yaml.instances;
         this.webhook = yaml.webhook;
         this.userIdPrefix = yaml.userIdPrefix || "_gitlab_";
+
+        for (const name in this.instances) {
+            const url = this.instances[name].url;
+            if (url.endsWith("/")) {
+                this.instances[name].url = url.slice(0, -1);
+            }
+        }
     }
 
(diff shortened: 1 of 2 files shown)


src/Config/Config.ts
```diff
@@ -209,6 +209,13 @@ export class BridgeConfigGitLab {
         this.instances = yaml.instances;
         this.webhook = yaml.webhook;
         this.userIdPrefix = yaml.userIdPrefix || "_gitlab_";
+
+        for (const name in this.instances) {
```
Half-Shot, line 213: This is fine, but we've generally moved to using `URL` for configuration these days.

PR #565, opened 2022-11-03
Harden against unauthorized changes to room state

AndrewFerr:
If an unauthorized change to connection-related room state is detected, deny the new connection settings & attempt to revert the state change.

Also catch an otherwise-uncaught error on authentication failure.

### DIFF
--- a/src/ConnectionManager.ts
+++ b/src/ConnectionManager.ts
@@ -91,7 +91,7 @@ export class ConnectionManager extends EventEmitter {
         throw new ApiError(`Connection type not known`);
     }
 
-    public assertStateAllowed(roomId: string, state: StateEvent<any>, serviceType: string) {
+    public assertStateAllowed(roomId: string, state: StateEvent, serviceType: string) {
         if (!this.isStateAllowed(roomId, state, serviceType)) {
             this.tryRestoreState(roomId, state, serviceType);
             log.error(`User ${state.sender} is disallowed to manage state for ${serviceType} in ${roomId}`);
@@ -101,12 +101,12 @@ export class ConnectionManager extends EventEmitter {
         }
     }
 
-    private isStateAllowed(roomId: string, state: StateEvent<any>, serviceType: string) {
+    private isStateAllowed(roomId: string, state: StateEvent, serviceType: string) {
         return state.sender === this.as.botUserId
             || this.config.checkPermission(state.sender, serviceType, BridgePermissionLevel.manageConnections);
     }
 
-    private async tryRestoreState(roomId: string, originalState: StateEvent<any>, serviceType: string) {
+    private async tryRestoreState(roomId: string, originalState: StateEvent, serviceType: string) {
         let state = originalState;
         try {
             do {
--- a/src/ConnectionManager.ts
+++ b/src/ConnectionManager.ts
@@ -114,7 +114,8 @@ export class ConnectionManager extends EventEmitter {
                 if (state.unsigned.replaces_state) {
                     state = new StateEvent(await this.as.botClient.getEvent(roomId, state.unsigned.replaces_state));
                 } else {
-                    await this.as.botClient.redactEvent(roomId, originalState.eventId);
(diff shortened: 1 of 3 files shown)


src/ConnectionManager.ts
```diff
@@ -91,12 +91,35 @@ export class ConnectionManager extends EventEmitter {
         throw new ApiError(`Connection type not known`);
     }
 
-    private assertStateAllowed(state: StateEvent<any>, serviceType: string) {
-        if (state.sender === this.as.botUserId) {
-            return;
+    public assertStateAllowed(roomId: string, state: StateEvent, serviceType: string) {
+        if (!this.isStateAllowed(roomId, state, serviceType)) {
+            this.tryRestoreState(roomId, state, serviceType);
```
jaller94, line 96: Personal preference: Tell the next person that you don't want to `catch` errors of an async function.

```suggestion
            void this.tryRestoreState(roomId, state, serviceType);
```

jaller94 (changes requested):
I'm asking you to write a back-off counter to avoid infinite or near-infinite loops.

src/ConnectionManager.ts
AndrewFerr: Ah, I didn't know about [the `void` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void), thanks!

src/ConnectionManager.ts
```diff
@@ -91,12 +91,35 @@ export class ConnectionManager extends EventEmitter {
...
+        let state = originalState;
+        try {
+            do {
+                if (state.unsigned.replaces_state) {
+                    state = new StateEvent(await this.as.botClient.getEvent(roomId, state.unsigned.replaces_state));
+                } else {
+                    await this.as.botClient.redactEvent(roomId, originalState.eventId);
+                    return;
+                }
+            } while (!this.isStateAllowed(roomId, state, serviceType));
```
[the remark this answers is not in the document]
AndrewFerr: Good idea.  I set the limit to 5 events.

AndrewFerr:
@jaller94 If the latest changes are fine by you, I'll merge this today.

Half-Shot (changes requested):
Apparently I failed to submit this review.

src/ConnectionManager.ts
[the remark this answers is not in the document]
AndrewFerr: True; I'll change the function name to make its purpose more clear.

src/ConnectionManager.ts
```diff
@@ -91,12 +91,56 @@ export class ConnectionManager extends EventEmitter {
...
+
+    private async tryRestoreState(roomId: string, originalState: StateEvent, serviceType: string) {
+        let state = originalState;
+        let attemptsRemaining = 5;
+        try {
+            do {
+                if (state.unsigned.replaces_state) {
+                    state = new StateEvent(await this.as.botClient.getEvent(roomId, state.unsigned.replaces_state));
+                } else {
+                    await this.as.botClient.redactEvent(roomId, originalState.eventId,
```
Half-Shot, line 136: I have worries this might cause a problem *if* you misconfigure and restart your hookshot, maybe this should only apply after restart?

Half-Shot (changes requested):
One concern, otherwise good!

src/ConnectionManager.ts
AndrewFerr: Sounds reasonable.  Then how should a disallowed state event be handled on startup?  Since the goal is to guard against a bad config redacting valid state events, should Hookshot just fail to start?

Half-Shot: Disallowed -> Ignore the connection entirely.

AndrewFerr: This is now addressed in 1bc4c25 by not rolling back invalid state events on startup.

Half-Shot (approved):

PR #573, opened 2022-11-17
Implement separate bot users per service

### LINKED ISSUE #436: Allow creation of "service bots" which can be used for specific service(s)
E.g. In the config you could specify you would like to have a JIRA bot that allows users to invite it to rooms in order to bridge JIRA things, without necessarily including all the other services too.

I'd probably advise having a new config for this like:

```yml
service_bots:
    - localpart: jira
      displayname: Jira
      avatar_url: mxc://jira
      services: [jira]
```

The bots could also create a widget for one specific type of configuration when prompted, which is issue https://github.com/matrix-org/matrix-hookshot/issues/437.

justinbot:
Introduces the ability for Hookshot to present itself as multiple different bots, each of which can handle one or more specific services. This is a pretty fundamental change, and as a result touches almost everywhere.

It should be relatively reviewable by commit, but will require significant testing of different scenarios to make sure things work correctly.

### Implementation
Fundamentally, this is breaking an assumption throughout the codebase that there is one global bot. The implications are:
  * Before interacting with a room, we need to find a bot which is in the room.
      * But also, multiple bots could be in the same room.
  * When doing so in the context of a particular service (e.g. sending a feed update), we need to prioritize the service bot over the default bot.

#### Connections
However, this does not mean connections are tied to a specific bot user.
We need to choose the right bot for a connection, but this is done at runtime based on which bots are in the room (again prioritizing service bots).

Fixes #436


src/Connections/FeedConnection.ts
```diff
@@ -42,13 +42,13 @@ const MAX_LAST_RESULT_ITEMS = 5;
 export class FeedConnection extends BaseConnection implements IConnection {
     static readonly CanonicalEventType = "uk.half-shot.matrix-hookshot.feed";
     static readonly EventTypes = [ FeedConnection.CanonicalEventType ];
-    static readonly ServiceCategory = "feed";
+    static readonly ServiceCategory = "feeds";
```
justinbot, line 45: The name used for this service was inconsistent -- I went with `feeds`, except for in config kept it `feed` to not break compatibility

src/Bridge.ts
```diff
@@ -776,21 +813,31 @@ export class Bridge {
...
+        const invitedUserId = event.state_key;
+        if (!invitedUserId) {
+            return;
+        }
+        log.info(`Got invite roomId=${roomId} from=${event.sender} to=${invitedUserId}`);
+
+        if (!this.botUsersManager.isBotUser(invitedUserId)) {
+            // We got an invite but it's not a configured bot user, must be for a ghost user
+            const client = this.as.getIntentForUserId(invitedUserId).underlyingClient;
+            return client.kickUser(invitedUserId, roomId, "Bridge does not support DMing ghosts");
```
justinbot, line 825: Would like a closer look here. I *think* the previous behavior was broken, and this would be correct (invited ghost user kicks itself).

src/Bridge.ts
```diff
@@ -934,19 +1008,22 @@ export class Bridge {
...
             return;
         }
 
-        // Only fetch rooms we have no connections in yet.
-        if (!this.connectionManager.isRoomConnected(roomId)) {
-            await this.connectionManager.createConnectionsForRoomId(roomId);
-        }
+        // Recreate connections for the room
+        await this.connectionManager.removeConnectionsForRoom(roomId);
+        await this.connectionManager.createConnectionsForRoomId(roomId);
```
justinbot, line 1026: This is intended to "hand off" connections to the correct bot, by deleting and recreating them. There might be a more elegant way to do it though.

src/Bridge.ts
```diff
@@ -817,16 +870,24 @@ export class Bridge {
...
+
+        if (!this.connectionManager) {
+            return;
+        }
+
+        // Remove all the connections for this room
+        await this.connectionManager.removeConnectionsForRoom(roomId);
+        if (this.joinedRoomsManager.getBotsInRoom(roomId).length > 0) {
+            // If there are still bots in the room, recreate connections
+            await this.connectionManager.createConnectionsForRoomId(roomId);
```
justinbot, line 890: This is intended to "hand off" connections to the correct bot, by deleting and recreating them (if we still have a bot in the room).

src/Bridge.ts
```diff
@@ -776,21 +823,40 @@ export class Bridge {
...
 
         // Don't accept invites from people who can't do anything
         if (!this.config.checkPermissionAny(event.sender, BridgePermissionLevel.login)) {
-            return this.as.botClient.kickUser(this.as.botUserId, roomId, "You do not have permission to invite this bot.");
+            return intent.underlyingClient.kickUser(invitedUserId, roomId, "You do not have permission to invite this bot.");
+        }
+
+        if (event.content.is_direct && invitedUserId !== this.as.botUserId) {
+            // Service bots do not support direct messages (admin rooms)
+            return intent.underlyingClient.kickUser(invitedUserId, roomId, "This bot does not support admin rooms.");
```
justinbot, line 847: This is mostly because of the additional complexity to determine which admin room commands each bot should handle.
If we want this, it can be added in a follow-up

src/Bridge.ts
```diff
@@ -934,19 +1029,22 @@ export class Bridge {
...
             return;
         }
 
-        // Only fetch rooms we have no connections in yet.
-        if (!this.connectionManager.isRoomConnected(roomId)) {
-            await this.connectionManager.createConnectionsForRoomId(roomId, true);
-        }
+        // Recreate connections for the room
+        await this.connectionManager.removeConnectionsForRoom(roomId);
+        await this.connectionManager.createConnectionsForRoomId(roomId, true);
```
justinbot, line 1047: This is intended to "hand off" connections to the correct bot, by deleting and recreating them

src/Bridge.ts
```diff
@@ -817,16 +890,24 @@ export class Bridge {
...
+
+        if (!this.connectionManager) {
+            return;
+        }
+
+        // Remove all the connections for this room
+        await this.connectionManager.removeConnectionsForRoom(roomId);
+        if (this.joinedRoomsManager.getBotsInRoom(roomId).length > 0) {
+            // If there are still bots in the room, recreate connections
+            await this.connectionManager.createConnectionsForRoomId(roomId, true);
```
justinbot, line 910: This is intended to "hand off" connections to the correct bot, by deleting and recreating them (if we still have a bot in the room)

src/Bridge.ts
```diff
@@ -776,21 +823,41 @@ export class Bridge {
...
+
         // Don't accept invites from people who can't do anything
         if (!this.config.checkPermissionAny(event.sender, BridgePermissionLevel.login)) {
-            return this.as.botClient.kickUser(this.as.botUserId, roomId, "You do not have permission to invite this bot.");
+            return intent.underlyingClient.kickUser(botUser.userId, roomId, "You do not have permission to invite this bot.");
+        }
+
+        if (event.content.is_direct && botUser.userId !== this.as.botUserId) {
+            // Service bots do not support direct messages (admin rooms)
+            return intent.underlyingClient.kickUser(botUser.userId, roomId, "This bot does not support admin rooms.");
```
justinbot, line 848: This is mostly because of the additional complexity to determine which admin room commands each bot should handle.
If we want this, it can be added in a follow-up

src/Bridge.ts
```diff
@@ -92,20 +110,45 @@ export class Bridge {
...
+            const intent = this.as.getIntentForUserId(botUser.userId);
+            const joinedRooms = await retry(() => intent.underlyingClient.getJoinedRooms(), 3, 3000);
+            log.debug(`Bot "${botUser.userId}" is joined to ${joinedRooms.length} rooms`);
+
+            for (const r of joinedRooms) {
+                this.joinedRoomsManager.addJoinedRoom(r, botUser.userId);
+            }
+        }
+
         await this.config.prefillMembershipCache(this.as.botClient);
```
justinbot, line 152: I'm pretty sure this needs changing to consider service bots, but I'm having trouble understanding where `bridgePermissions.getInterestedRooms` comes from.

src/Managers/JoinedRoomsManager.ts
```diff
@@ -0,0 +1,63 @@
+export default class JoinedRoomsManager {
+    // Map of room ID to set of our bot user IDs in the room
```
Half-Shot, line 2: I think this class logic could b be merged into `BotUsersManager`? It seems quite intertwined.

src/Bridge.ts
```diff
@@ -758,36 +803,59 @@ export class Bridge {
...
-        // Room joins can fail over federation
-        if (event.state_key !== this.as.botUserId) {
-            return this.as.botClient.kickUser(event.state_key, roomId, "Bridge does not support DMing ghosts");
+        const invitedUserId = event.state_key;
+        if (!invitedUserId) {
+            return;
+        }
+        log.info(`Got invite roomId=${roomId} from=${event.sender} to=${invitedUserId}`);
+
+        const botUser = this.botUsersManager.getBotUser(invitedUserId);
```
Half-Shot, line 812: Could getBotUser return an object that returns an intent, to save on LoC.

Half-Shot, line 812: Could we also be explicit and call this a serviceBotUser so it doesn't confuse with the sender_localpart user.

src/Bridge.ts
Half-Shot: This should probably be a dedicated handover function. Our connections are slightly stateful and it's better to not recreate them v.s. updating them, if we can avoid it.

src/Bridge.ts
```diff
@@ -833,26 +901,40 @@ export class Bridge {
...
                 // Divert to the setup room code if we didn't match any of these
-                try {
-                    await (
-                        new SetupConnection(
+
+                const botUsersInRoom = this.botUsersManager.getBotUsersInRoom(roomId);
+                // Try each bot in the room until one handles the command
+                for (const botUser of botUsersInRoom) {
+                    try {
+                        const setupConnection = new SetupConnection(
```
Half-Shot, line 909: Could setup connection instead take the `botUsersInRoom` and iterate within the connection, rather than recreating the connection in a loop?

Half-Shot (changes requested):
Bunch of comments, @tadzik has more :)

src/Bridge.ts
```diff
@@ -289,108 +346,108 @@ export class Bridge {
 
         this.bindHandlerToQueue<GitHubWebhookTypes.IssuesUnlabeledEvent, GitHubRepoConnection>(
             "github.issues.unlabeled",
-            (data) => connManager.getConnectionsForGithubRepo(data.repository.owner.login, data.repository.name), 
+            (data) => connManager.getConnectionsForGithubRepo(data.repository.owner.login, data.repository.name),
```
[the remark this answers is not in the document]
justinbot: My IDE takes care of trailing whitespace automatically but a rule would be nice in case it sneaks in

src/Bridge.ts
justinbot: This is not specific to service bots, `getBotUser` abstracts over service bots and the default bot.

justinbot: Good idea

src/Bridge.ts
justinbot: That seems tricky and would break the paradigm where connections are instantiated with a specific intent (i.e. what bot client would we pass to the `CommandConnection` parent class?)

I thought this was a fairly elegant use of connections, unless there is a strong reason it's not good to instantiate them repeatedly?

src/Bridge.ts
justinbot: Done, not sure how I wrote that so many times without thinking of a better way 😄

src/Config/Config.ts
```diff
@@ -471,6 +480,8 @@ export class BridgeConfig {
     public readonly feeds?: BridgeConfigFeeds;
     @configKey("Define profile information for the bot user", true)
     public readonly bot?: BridgeConfigBot;
+    @configKey("EXPERIMENTAL Define additional bot users for specific services", true)
```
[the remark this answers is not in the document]
justinbot: Not any more so than usual I suppose 😄

src/Bridge.ts
```diff
@@ -57,6 +62,7 @@ export class Bridge {
     constructor(
         private config: BridgeConfig,
         private readonly listener: ListenerService,
+        private readonly registration: IAppserviceRegistration,
```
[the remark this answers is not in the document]
justinbot: Took me a second to realize what BUM meant 🤣 but yes I agree

src/Bridge.ts
```diff
@@ -983,28 +1086,31 @@ export class Bridge {
...
-                        log.error(`Failed to create setup widget for ${roomId}`, ex);
+            const botUsersInRoom = this.botUsersManager.getBotUsersInRoom(roomId);
+            for (const botUser of botUsersInRoom) {
+                // If it's a power level event for a new room, we might want to create the setup widget.
+                if (this.config.widgets?.roomSetupWidget?.addOnInvite && event.type === "m.room.power_levels" && event.state_key === "" && !this.connectionManager.isRoomConnected(roomId)) {
+                    log.debug(`${roomId} got a new powerlevel change and isn't connected to any connections, testing to see if we should create a setup widget`)
+                    const plEvent = new PowerLevelsEvent(event);
+                    const currentPl = plEvent.content.users?.[botUser.userId] || plEvent.defaultUserLevel;
+                    const previousPl = plEvent.previousContent?.users?.[botUser.userId] || plEvent.previousContent?.users_default;
+                    const requiredPl = plEvent.content.events?.["im.vector.modular.widgets"] || plEvent.defaultStateEventLevel;
```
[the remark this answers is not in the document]
justinbot: This should be fixed separately as it's not strictly related to this PR

src/Connections/GenericHook.ts
```diff
@@ -200,52 +198,50 @@ export class GenericHookConnection extends BaseConnection implements IConnection
...
+        private readonly intent: Intent,
+    ) {
+        super(roomId, stateKey, GenericHookConnection.CanonicalEventType);
+        if (state.transformationFunction && config.allowJsTransformationFunctions) {
+            this.transformationFunction = new Script(state.transformationFunction);
         }
+    }
 
+    public get priority(): number {
+        return this.state.priority || super.priority;
```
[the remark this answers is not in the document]
justinbot: This should be fixed separately as it's not strictly related to this PR

src/Connections/GenericHook.ts
```diff
@@ -200,52 +198,50 @@ export class GenericHookConnection extends BaseConnection implements IConnection
...
         return GenericHookConnection.EventTypes.includes(eventType) && this.stateKey === stateKey;
     }
 
     public getUserId() {
         if (!this.config.userIdPrefix) {
-            return this.as.botUserId;
+            return this.intent.userId;
         }
-        const [, domain] = this.as.botUserId.split(':');
+        const [, domain] = this.intent.userId.split(':');
```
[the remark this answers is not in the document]
justinbot: This should be fixed separately as it's not strictly related to this PR

src/Widgets/SetupWidget.ts
```diff
@@ -16,15 +16,24 @@ export class SetupWidget {
         return false;
     }
 
-    static async SetupRoomConfigWidget(roomId: string, botIntent: Intent, config: BridgeWidgetConfig): Promise<boolean> {
-        if (await SetupWidget.createWidgetInRoom(roomId, botIntent, config, HookshotWidgetKind.RoomConfiguration, "hookshot_room_config")) {
+    static async SetupRoomConfigWidget(roomId: string, botIntent: Intent, config: BridgeWidgetConfig, serviceTypes: string[]): Promise<boolean> {
+        // If this is for a single service, scope the widget
+        const serviceScope = serviceTypes.length === 1 ? serviceTypes[0] : undefined;
+        if (await SetupWidget.createWidgetInRoom(roomId, botIntent, config, HookshotWidgetKind.RoomConfiguration, "hookshot_room_config", serviceScope)) {
```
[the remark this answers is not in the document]
justinbot: Good point -- addressed this with state keys as suggested.
Also set the `serviceScope` as the widget subtitle so you can differentiate them without opening.

src/Connections/GenericHook.ts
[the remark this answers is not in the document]
jaller94: PR #600

src/Connections/FigmaFileConnection.ts
```diff
@@ -100,16 +101,21 @@ export class FigmaFileConnection extends BaseConnection implements IConnection {
             return;
         }
 
-        const intent = this.as.getIntentForUserId(this.config.overrideUserId || this.as.botUserId);
-    
+        let intent;
+        if (this.config.overrideUserId) {
```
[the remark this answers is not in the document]
Half-Shot: Can we gracefully assume that if overrideUserId is set in the config, we should warn about deprecation and create a service bot for it?

If the service bot config is *also*  defined, then skip it and fail if it's still defined in the config.

src/Bridge.ts
Half-Shot: I think we decided this was fine in the end.

Half-Shot (approved):

src/Bridge.ts
Half-Shot: I think this is fine, we can use the lead bot for this one.

PR #578, opened 2022-11-25
Save bot-created admin rooms in profile data

AndrewFerr:
This prevents a process restart from causing bot-created admin rooms to be forgotten.

### DIFF
--- a/changelog.d/578.bugfix
+++ b/changelog.d/578.bugfix
@@ -1,1 +1,1 @@
-Track bot-created admin rooms in the bot's room profile data, not just user-created admin rooms.
+Fixed a case where a bridge-created admin room would stop working on restart.
--- a/changelog.d/578.bugfix
+++ b/changelog.d/578.bugfix
@@ -1,0 +1,1 @@
+Track bot-created admin rooms in the bot's room profile data, not just user-created admin rooms.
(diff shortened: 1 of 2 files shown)


changelog.d/578.bugfix
```diff
@@ -0,0 +1 @@
+Track bot-created admin rooms in the bot's room profile data, not just user-created admin rooms.
```
Half-Shot, line 1: What's the user-facing change there? What bug are we fixing?

Half-Shot (changes requested):

changelog.d/578.bugfix
AndrewFerr: To reproduce:
- invite hookshot to a room & give it state permissions
- run `!hookshot webhook <name>`
- get invited to an admin room DM by hookshot
- reboot the bridge
- run command in the admin room DM that hookshot created for you

Result: Hookshot will ignore commands in that admin room, because it forgot that it is one.

Half-Shot: Ah yeah now I remember this. I'd just say "Fixed a case where a bridge-created admin room would stop working on restart"

AndrewFerr: Done :)

Half-Shot (approved):
Great stuff, thank you for fixing these cases!

PR #587, opened 2022-12-07
Ensure unknown keys are retained when connection state is altered via widget

### LINKED ISSUE #584: Updating connection state via widget after manual edits, will obliterate manual edits.
I think this is probably unintentional, as we want to retain unknown keys.

Half-Shot:
Fixes #584 

This also fixes a few validation bugs in GitHubRepo.


src/Connections/GithubRepo.ts
```diff
@@ -51,7 +51,7 @@ export interface GitHubRepoConnectionOptions extends IConnectionState {
     excludingLabels?: string[];
     hotlinkIssues?: boolean|{
         prefix: string;
-    };
+    }|null;
```
Half-Shot, line 54: Remove this

PR #592, opened 2022-12-09
Drop `ignoreHooks` configuration.

### LINKED ISSUE #591: Drop ignoredHooks state property, only support enabledHooks
Instead, new connections will start with a default set of hooks in the state. Old connections will be automatically upgraded to the new system as a once-off.

Half-Shot:
Fixes #591 

This is a bit of a large refactor, but should massively simplify our event system overall. This:

- Refactors the backend to *only* support *either* `enabledHooks` or defaultHooks.
- Refactors the frontend to support that.
- Refactors a bit of the frontend to split out shared components.
- Cleans up our jsx/preact configuration.

This does NOT autoupgrade state, as that's fairly error prone and we don't want to forcibly do it on startup. This instead adopts an approach of consuming the legacy `ignoreHooks` field from existing state, while overwriting existing state with new keys automatically.

### DIFF
--- a/docs/usage/room_configuration/github_repo.md
+++ b/docs/usage/room_configuration/github_repo.md
@@ -27,8 +27,8 @@ This connection supports a few options which can be defined in the room state:
 
 | Option | Description | Allowed values | Default |
 |--------|-------------|----------------|---------|
-|enableHooks [^1]|Enable notifications for some event types|Array of: [Supported event types](#supported-event-types) |*empty*|
-|ignoreHooks [^1]|Choose to exclude notifications for some event types|Array of: [Supported event types](#supported-event-types) |*empty*|
+|enableHooks [^1]|Enable notifications for some event types|Array of: [Supported event types](#supported-event-types) |If not defined, defaults are mentioned below|
+|ignoreHooks [^1]|**deprecated** Choose to exclude notifications for some event types|Array of: [Supported event types](#supported-event-types) |*empty*|
 |commandPrefix|Choose the prefix to use when sending commands to the bot|A string, ideally starts with "!"|`!gh`|
 |showIssueRoomLink|When new issues are created, provide a Matrix alias link to the issue room|`true/false`|`false`|
 |prDiff|Show a diff in the room when a PR is created, subject to limits|`{enabled: boolean, maxLines: number}`|`{enabled: false}`|
@@ -43,7 +43,7 @@ This connection supports a few options which can be defined in the room state:
 |workflowRun.excludingWorkflows|Never report workflow runs with a matching workflow name.|Array of: String matching a workflow name|*empty*|
 
 
-[^1]: `ignoreHooks` takes precedence over `enableHooks`.
+[^1]: `ignoreHooks` is no longer accepted for new state events. Use `enableHooks` to explicitly state all events you want to see.
 
 
 ### Supported event types
--- a/docs/usage/room_configuration/github_repo.md
+++ b/docs/usage/room_configuration/github_repo.md
@@ -46,11 +46,13 @@ This connection supports a few options which can be defined in the room state:
 [^1]: `ignoreHooks` is no longer accepted for new state events. Use `enableHooks` to explicitly state all events you want to see.
 
 
+
 ### Supported event types
 
 This connection supports sending messages when the following actions happen on the repository.
 
-Note: Some of these event types are enabled by default (marked with a `*`)
+Note: Some of these event types are enabled by default (marked with a `*`). When `ignoreHooks` *is* defined,
+the events marked as default below will be enabled. Otherwise, this is ignored.
 
 - issue *
   - issue.created *
--- a/src/Connections/GithubRepo.ts
+++ b/src/Connections/GithubRepo.ts
@@ -340,6 +340,7 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
                 }
                 log.warn(`Room has old state key 'ignoreHooks'. Converting to compatible enabledHooks filter`);
                 state.enableHooks = HookFilter.convertIgnoredHooksToEnabledHooks(state.enableHooks, state.ignoreHooks, DefaultHooks);
+                delete state.ignoreHooks;
             }
             return state;
         }
--- a/src/Connections/GithubRepo.ts
+++ b/src/Connections/GithubRepo.ts
@@ -40,8 +40,12 @@ interface IQueryRoomOpts {
 }
 
 export interface GitHubRepoConnectionOptions extends IConnectionState {
-    enableHooks?: AllowedEventsNames[],
+    /**
+     * Do not use. Use `enableHooks`.
+     * @deprecated
+     */
     ignoreHooks?: AllowedEventsNames[],
+    enableHooks?: AllowedEventsNames[],
     showIssueRoomLink?: boolean;
     prDiff?: {
         enabled: boolean;
@@ -136,8 +140,17 @@ const AllowedEvents: AllowedEventsNames[] = [
  * These hooks are enabled by default, unless they are
  * specifed in the ignoreHooks option.
  */
-const AllowHookByDefault: AllowedEventsNames[] = [
+const DefaultHooks: AllowedEventsNames[] = [
+    "issue.changed",
+    "issue.created",
+    "issue.edited",
+    "issue.labeled",
     "issue",
+    "pull_request.closed",
+    "pull_request.merged",
+    "pull_request.opened",
+    "pull_request.ready_for_review",
+    "pull_request.reviewed",
     "pull_request",
     "release.created"
 ];
@@ -151,6 +164,10 @@ const ConnectionStateSchema = {
     },
     org: {type: "string"},
     repo: {type: "string"},
+    /**
+     * Legacy state.
+     * @deprecated
+     */
     ignoreHooks: {
         type: "array",
         items: {
@@ -311,16 +328,18 @@ export interface GitHubTargetFilter {
  */
 @Connection
 export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnectionState> implements IConnection {
-
 	static validateState(state: unknown, isExistingState = false): GitHubRepoConnectionState {
         const validator = new Ajv({ allowUnionTypes: true }).compile(ConnectionStateSchema);
         if (validator(state)) {
-            // Validate ignoreHooks IF this is an incoming update (we can be less strict for existing state)
-            if (!isExistingState && state.ignoreHooks && !state.ignoreHooks.every(h => AllowedEvents.includes(h))) {
-                throw new ApiError('`ignoreHooks` must only contain allowed values', ErrCode.BadValue);
-            }
             if (!isExistingState && state.enableHooks && !state.enableHooks.every(h => AllowedEvents.includes(h))) {
                 throw new ApiError('`enableHooks` must only contain allowed values', ErrCode.BadValue);
+            }
+            if (state.ignoreHooks) {
+                if (!isExistingState) {
+                    throw new ApiError('`ignoreHooks` cannot be used with new connections', ErrCode.BadValue);
+                }
+                log.warn(`Room has old state key 'ignoreHooks'. Converting to compatible enabledHooks filter`);
+                state.enableHooks = HookFilter.convertIgnoredHooksToEnabledHooks(state.enableHooks, state.ignoreHooks, DefaultHooks);
             }
             return state;
         }
@@ -489,9 +508,7 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
                 "github",
             );
         this.hookFilter = new HookFilter(
-            AllowHookByDefault,
-            state.enableHooks,
-            state.ignoreHooks,
+            state.enableHooks ?? DefaultHooks,
         )
     }
 
@@ -531,7 +548,6 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
     public async onStateUpdate(stateEv: MatrixEvent<unknown>) {
         await super.onStateUpdate(stateEv);
         this.hookFilter.enabledHooks = this.state.enableHooks ?? [];
-        this.hookFilter.ignoredHooks = this.state.ignoreHooks ?? [];
     }
 
     public isInterestedInStateEvent(eventType: string, stateKey: string) {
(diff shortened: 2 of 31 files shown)


docs/usage/room_configuration/github_repo.md
```diff
@@ -43,7 +43,7 @@ This connection supports a few options which can be defined in the room state:
 |workflowRun.excludingWorkflows|Never report workflow runs with a matching workflow name.|Array of: String matching a workflow name|*empty*|
 
 
-[^1]: `ignoreHooks` takes precedence over `enableHooks`.
+[^1]: `ignoreHooks` is no longer accepted for new state events. Use `enableHooks` to explicitly state all events you want to see.
```
Half-Shot, line 46: TODO: Maybe clarify default behaviours

src/Connections/GithubRepo.ts
```diff
@@ -40,8 +40,12 @@ interface IQueryRoomOpts {
 }
```
Half-Shot, line 40: TODO: Check Jira

src/HookFilter.ts
```diff
@@ -1,19 +1,32 @@
 export class HookFilter<T extends string> {
+    static convertIgnoredHooksToEnabledHooks<T extends string>(enabledHooks: T[] = [], ignoredHooks: T[] = [], defaultHooks: T[]): T[] {
```
Half-Shot, line 2: Naming is awkward.

src/HookFilter.ts
```diff
@@ -1,19 +1,32 @@
...
+
     constructor(
-        public readonly defaultHooks: T[],
         public enabledHooks: T[] = [],
-        public ignoredHooks: T[] = []
     ) {
 
     }
 
     public shouldSkip(...hookName: T[]) {
```
Half-Shot, line 29: Eventually rename this to be positive

tests/HookFilter.ts
```diff
@@ -3,29 +3,56 @@ import { HookFilter } from '../src/HookFilter';
...
+    
+    describe('convertIgnoredHooksToEnabledHooks', () => {
+        it('should correctly provide a list of default hooks', () => {
+            expect(HookFilter.convertIgnoredHooksToEnabledHooks([], [], DEFAULT_SET)).to.deep.equal(
+                DEFAULT_SET
+            );
+        });
+        
+        it('should correctly include both default and enabled hooks', () => {
+            expect(HookFilter.convertIgnoredHooksToEnabledHooks(ENABLED_SET, [], DEFAULT_SET)).to.deep.equal([
```
Half-Shot, line 29: Don't assume ordering.

tests/HookFilter.ts
```diff
@@ -3,29 +3,56 @@ import { HookFilter } from '../src/HookFilter';
...
-        expect(filter.shouldSkip('enabled-hook', 'enabled-but-ignored')).to.be.true;
+    
+    describe('convertIgnoredHooksToEnabledHooks', () => {
+        it('should correctly provide a list of default hooks', () => {
+            expect(HookFilter.convertIgnoredHooksToEnabledHooks([], [], DEFAULT_SET)).to.deep.equal(
+                DEFAULT_SET
+            );
+        });
+        
+        it('should correctly include both default and enabled hooks', () => {
```
Half-Shot, line 28: This should only return the enabled set since defaults only come into play with ignored hooks.

tests/HookFilter.ts
```diff
@@ -3,29 +3,56 @@ import { HookFilter } from '../src/HookFilter';
...
+        
+        it('should deduplicate', () => {
+            expect(HookFilter.convertIgnoredHooksToEnabledHooks(DEFAULT_SET, [], DEFAULT_SET)).to.deep.equal([
+                ...DEFAULT_SET
+            ]);
+        });
+        
+        it('should correctly exclude ignored hooks', () => {
+            expect(HookFilter.convertIgnoredHooksToEnabledHooks([], [DEFAULT_SET[0]], DEFAULT_SET)).to.deep.equal([
+                DEFAULT_SET[1]
```
Half-Shot, line 42: Just filter out DEFAULT_SET[0]

tests/HookFilter.ts
```diff
@@ -3,29 +3,56 @@ import { HookFilter } from '../src/HookFilter';
...
+            ]);
+        });
+        
+        it('should handle ignored root hooks', () => {
+            const defaultHooks = ['myhook', 'myhook.foo', 'myhook.foo.bar'];
+            expect(HookFilter.convertIgnoredHooksToEnabledHooks([], [], defaultHooks)).to.deep.equal(defaultHooks);
```
Half-Shot, line 48: Redundant, no ignored hooks.

tests/HookFilter.ts
```diff
@@ -3,29 +3,56 @@ import { HookFilter } from '../src/HookFilter';
...
+            expect(HookFilter.convertIgnoredHooksToEnabledHooks([], [defaultHooks[2]], defaultHooks)).to.deep.equal([
```
Half-Shot, line 49: Just use the names, easier to read.

src/Connections/GithubRepo.ts
```diff
@@ -489,9 +509,7 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
                 "github",
             );
         this.hookFilter = new HookFilter(
-            AllowHookByDefault,
-            state.enableHooks,
-            state.ignoreHooks,
+            state.enableHooks ?? DefaultHooks,
```
Half-Shot, line 512: We need this, otherwise a lack of defaults on missing state would be a breaking change.

src/Connections/IConnection.ts
```diff
@@ -81,7 +81,8 @@ export interface ConnectionDeclaration<C extends IConnection = IConnection> {
     EventTypes: string[];
     ServiceCategory: string;
     provisionConnection?: (roomId: string, userId: string, data: Record<string, unknown>, opts: ProvisionConnectionOpts) => Promise<{connection: C, warning?: ConnectionWarning}>;
-    createConnectionForState: (roomId: string, state: StateEvent<Record<string, unknown>>, opts: InstantiateConnectionOpts) => C|Promise<C>
+    checkIfStateRequiresUpgrade?: (data: StateEvent<Record<string, unknown>>) => Record<string, unknown>|false;
```
Half-Shot, line 84: Never got used. Please remove.
```suggestion
```

tests/connections/GithubRepoTest.ts
```diff
@@ -93,24 +104,24 @@ describe("GitHubRepoConnection", () => {
...
 			}
 		});
-		it("will disallow ignoreHooks to contains invalid enums if this is new state", () => {
+		it("will disallow enabledHooks to contains invalid enums if this is new state", () => {
 			try {
 				GitHubRepoConnection.validateState({
 					org: "foo",
 					repo: "bar",
-					ignoreHooks: ["issue", "pull_request", "release", "not-real"],
+					enabledHooks: ["issue", "pull_request", "release", "not-real"],
```
Half-Shot, line 112: Remove valid hooks to make it obvious.

tests/connections/GithubRepoTest.ts
```diff
@@ -93,24 +104,24 @@ describe("GitHubRepoConnection", () => {
...
 				}
 			}
 		});
-		it("will allow ignoreHooks to contains invalid enums if this is old state", () => {
+		it("will allow enabledHooks to contains invalid enums if this is old state", () => {
 			GitHubRepoConnection.validateState({
 				org: "foo",
 				repo: "bar",
-				ignoreHooks: ["issue", "pull_request", "release", "not-real"],
+				enabledHooks: ["issue", "pull_request", "release", "not-real"],
```
Half-Shot, line 124: Remove valid hooks to make it obvious.

tests/connections/GitlabRepoTest.ts
```diff
@@ -79,6 +80,17 @@ describe("GitLabRepoConnection", () => {
 				excludingLabels: ["but-not-me"],
 			} as GitLabRepoConnectionState as unknown as Record<string, unknown>);
 		});
+		it("will convert ignoredHooks for existing state", () => {
+			const state = GitLabRepoConnection.validateState({
+				instance: "foo",
```
Half-Shot, line 85: Add default correct state as const, then override.

web/components/elements/EventHookCheckbox.tsx
```diff
@@ -0,0 +1,26 @@
+import { FunctionComponent } from "preact";
+import { JSXInternal } from "preact/src/jsx";
+
+export const EventHookCheckbox: FunctionComponent<{
+    enabledHooks?: string[],
+    onChange: JSXInternal.GenericEventHandler<HTMLInputElement>,
+    eventName: string,
```
Half-Shot, line 7: Maybe call this hook name, or hookshothookname

src/Connections/GithubRepo.ts
Half-Shot: Jira effectively has `enableHooks`, except we just called it `events.

docs/usage/room_configuration/github_repo.md
Half-Shot: Done.

tests/HookFilter.ts
Half-Shot: `to.have.members` does this for us!

tests/HookFilter.ts
Half-Shot: Our assumptions were wrong in review. `convertIgnoredHooksToEnabledHooks` is only called when ignoreHooks is non-falsey, so we can assume that it will always have values. The test here is fine.

src/Connections/GithubRepo.ts
Half-Shot: Added stronger typing.

src/Connections/GitlabRepo.ts
```diff
@@ -152,14 +175,26 @@ export class GitLabRepoConnection extends CommandConnection<GitLabRepoConnection
...
+                throw new ApiError('`enableHooks` must only contain allowed values', ErrCode.BadValue);
             }
-            return state;
+            if (state.ignoreHooks) {
+                if (!isExistingState) {
+                    throw new ApiError('`ignoreHooks` cannot be used with new connections', ErrCode.BadValue);
+                }
+                log.warn(`Room has old state key 'ignoreHooks'. Converting to compatible enabledHooks filter`);
+                state.enableHooks = HookFilter.convertIgnoredHooksToEnabledHooks(state.enableHooks, state.ignoreHooks, AllowedEvents);
+                delete state.ignoreHooks;
```
jaller94, line 191: Isn't this a duplication of `ignoreHooks: undefined,` while also changing `state` as a side effect?

src/Connections/GithubRepo.ts
```diff
@@ -310,19 +333,26 @@ export interface GitHubTargetFilter {
...
                 throw new ApiError('`enableHooks` must only contain allowed values', ErrCode.BadValue);
             }
-            return state;
+            if (state.ignoreHooks) {
+                if (!isExistingState) {
+                    throw new ApiError('`ignoreHooks` cannot be used with new connections', ErrCode.BadValue);
+                }
+                log.warn(`Room has old state key 'ignoreHooks'. Converting to compatible enabledHooks filter`);
+                state.enableHooks = HookFilter.convertIgnoredHooksToEnabledHooks(state.enableHooks, state.ignoreHooks, DefaultHooks);
+                delete state.ignoreHooks;
```
jaller94, line 349: Isn't this a duplication of `ignoreHooks: undefined,` while also changing `state` as a side effect?

web/components/elements/EventHookCheckbox.tsx
```diff
@@ -0,0 +1,26 @@
...
+    parentEvent?: string,
+}> = ({enabledHooks, onChange, eventName, parentEvent, children}) => {
+    if (!enabledHooks) {
+        throw Error(`Invalid configuration for checkbox ${eventName}`);
```
jaller94, line 11: After 5 years of coding with React, this is the first component I see which intentionally `throw`s.
Are you sure you don't want to display something?

src/HookFilter.ts
```diff
@@ -1,19 +1,31 @@
...
+            ...defaultHooks,
+        ]);
+
+        // For each ignored hook, remove anything that matches.
+        for (const ignoredHook of ignoredHooks) {
+            resultHookSet.delete(ignoredHook);
+            // If the hook is a "root" hook name, remove all children.
+            for (const enabledHook of resultHookSet) {
+                if (enabledHook.startsWith(`${ignoredHook}.`)) {
+                    resultHookSet.delete(enabledHook);
```
jaller94, line 14: I looked up that deleting items while iterating over a Set (or array) is ok.
Seems to be safe, as long as you're only deleting the item of the current iteration. :+1:

src/HookFilter.ts
```diff
@@ -1,19 +1,31 @@
...
 
     public shouldSkip(...hookName: T[]) {
-        if (hookName.some(name => this.ignoredHooks.includes(name))) {
-            return true;
-        }
-        if (hookName.some(name => this.enabledHooks.includes(name))) {
-            return false;
-        }
-        return !hookName.some(h => this.defaultHooks.includes(h));
+        return !hookName.some(name => this.enabledHooks.includes(name));
```
jaller94, line 29: Speaking about positive naming:
Isn't `foo.every()` the same as `!foo.some()`?
```suggestion
        return hookName.every(name => this.enabledHooks.includes(name));
```

src/Connections/GithubRepo.ts
```diff
@@ -488,10 +518,9 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
                 "!gh",
                 "github",
             );
+            // It's possible to have state with no `enableHooks`, so we need
```
jaller94, line 521: Weird indentation.
Also, "so we need" seems to be an incomplete sentence.

src/Connections/GitlabRepo.ts
Half-Shot: Good point. Should remove that one.

web/components/elements/EventHookCheckbox.tsx
Half-Shot: Tbh, I think type checking is enough. I'll remove the error.

src/HookFilter.ts
Half-Shot: Aha, yes.

tests/connections/GitlabRepoTest.ts
Half-Shot: I think in the end this isn't quite noisy enough to warrant it, so I've left it for a future refactor.

PR #600, opened 2022-12-30
Support server names with colons (e.g. IPv6, custom port)

### LINKED ISSUE #597: Server names with colons (IPv6, non-default port) aren't supported
The code has assumptions that server names (e.g. in user IDs) don't contain colons.

However, according to the Matrix Spec, a server name may contain colons, as part of an IPv6 or defining a non-default port.
https://spec.matrix.org/v1.5/appendices/#server-name

## Code
https://github.com/matrix-org/matrix-hookshot/blob/c962f17a91ab923c4c2f17aa09b6feb287d377fa/src/Connections/GenericHook.ts#L229

There may be more places that need to be found, fixed and tested.

jaller94:
Fixes #597.

### DIFF
--- a/src/Connections/GenericHook.ts
+++ b/src/Connections/GenericHook.ts
@@ -226,10 +226,10 @@ export class GenericHookConnection extends BaseConnection implements IConnection
         if (!this.config.userIdPrefix) {
             return this.as.botUserId;
         }
-        const [, domain] = this.as.botUserId.split(':');
+        const serverName = this.as.botUserId.slice(this.as.botUserId.indexOf(":") + 1);
         const name = this.state.name &&
              this.state.name.replace(/[A-Z]/g, (s) => s.toLowerCase()).replace(/([^a-z0-9\-.=_]+)/g, '');
-        return `@${this.config.userIdPrefix}${name || 'bot'}:${domain}`;
+        return `@${this.config.userIdPrefix}${name || 'bot'}:${serverName}`;
     }
 
     public async ensureDisplayname(sender: string) {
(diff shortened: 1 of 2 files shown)


tests/config/permissions.ts
```diff
@@ -110,7 +132,7 @@ describe("Config/BridgePermissions", () => {
         const checkActorValues = ["@foo:bar", "bar", "*"];
         checkActorValues.forEach(actor => {
             it(`will return true for a service defintion of '${actor}' that has a sufficent level`, () => {
-                const bridgePermissions = genBridgePermissions("@foo:bar", "fake-service", "commands");
+                const bridgePermissions = genBridgePermissions(actor, "fake-service", "commands");
```
jaller94, line 135: Unrelated fix. The `forEach` wasn't using `actor` to generate the bridge permission, so "bar" and "*" were never tested.

src/Config/permissions.rs
```diff
@@ -106,17 +106,11 @@ impl BridgePermissions {
         service: String,
         permission: String,
     ) -> napi::Result<bool> {
-        let parts: Vec<&str> = mxid.split(':').collect();
-        let domain: String;
+        let server_name = mxid.split_once(":").unwrap().1.to_owned();
```
jaller94, line 109: Reivilibre came up with this one. :pray: 
Poljar and reivilibre recommended to use Ruma instead of re-writing a parser.

https://docs.rs/ruma-identifiers/latest/ruma_identifiers/struct.UserId.html

jaller94, line 109: Beware of this change:
The new `.unwrap()` will throw, if the `mxid` is invalid and doesn't contain a colon.

src/Config/permissions.rs
```diff
@@ -106,17 +106,11 @@ impl BridgePermissions {
...
         let permission_int = permission_level_to_int(permission)?;
-        if parts.len() > 1 {
```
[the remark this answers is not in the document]
jaller94: It was not ok to pass a domain. This is always expected to be a Matrix User ID.

I checked this and would have renamed `mxid` to `mxidOrServerName`.

PR #602, opened 2023-01-02
A11y: Add alt tags to all images

jaller94:
* `alt=""` indicates that there's no additional information in the image. It can be used for avatars and icons where the label is next to the image.

### DIFF
--- a/web/components/ConnectionCard.tsx
+++ b/web/components/ConnectionCard.tsx
@@ -11,7 +11,7 @@ interface IProps {
 
 export function ConnectionCard(props: IProps) {
     return <div className={style.card} onClick={props.onClick}>
-        <img src={props.imageSrc} />
+        <img alt="" src={props.imageSrc} />
         <div>
             <span>{props.serviceName}</span>
             <p>{props.description}</p>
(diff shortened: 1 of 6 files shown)


web/components/ConnectionCard.tsx
```diff
@@ -11,7 +11,7 @@ interface IProps {
 
 export function ConnectionCard(props: IProps) {
     return <div className={style.card} onClick={props.onClick}>
-        <img src={props.imageSrc} />
+        <img alt="" src={props.imageSrc} />
```
Half-Shot, line 14: TIL `alt=""` is valid. Should we still add a description like "Service logo"?

web/components/roomConfig/RoomConfig.tsx
```diff
@@ -100,7 +100,7 @@ export const RoomConfig = function<SConfig, ConnectionType extends GetConnection
                 )
         }
         <header className={style.header}>
-            <img src={headerImg} />
+            <img alt="" src={headerImg} />
```
Half-Shot, line 103:
```suggestion
            <img alt="Service logo" src={headerImg} />
```

web/components/ConnectionCard.tsx
Half-Shot, line 14:
```suggestion
        <img alt="Service logo" src={props.imageSrc} />
```

Half-Shot (approved):
Thanks for the continued focus on a11y!

web/components/ConnectionCard.tsx
jaller94: It has the serviceName printed right next to it.
And a user of a screen reader is unlikely to be interested in the presence of a logo.

PR #603, opened 2023-01-02
BotCommands: Don't use user input as JavaScript object keys

jaller94:


src/BotCommands.ts
```diff
@@ -38,42 +38,44 @@ export interface BotCommandOptions {
 type BotCommandResult = {status?: boolean, reaction?: string}|undefined;
 type BotCommandFunction = (...args: string[]) => Promise<BotCommandResult>;
 
-export type BotCommands = {[prefix: string]: {fn: BotCommandFunction} & BotCommandOptions};
+export type BotCommands = Map<string, {fn: BotCommandFunction} & BotCommandOptions>;
 export type HelpFunction = (cmdPrefix?: string, categories?: string[], includeTitles?: boolean) => MatrixMessageContent
 
 export function compileBotCommands(...prototypes: Record<string, BotCommandFunction>[]): {helpMessage: HelpFunction, botCommands: BotCommands} {
```
Half-Shot, line 44: I do wonder if we should add a test case for BotCommands to just ensure nothing breaks?

Half-Shot (changes requested):
Seems fine, but I hope we have adequate tests for BotCommands

src/BotCommands.ts
jaller94: I tried... but this relies on Reflect and decorators (not enabled for the tests director yet), multiple levels of inheritance and using prototypes as JavaScript objects.
All three on their own make code super hard to write unit tests for.

Half-Shot:
@jaller94 In the interests of time, I'll settle for just having you eyeball this and test that nothing is broken on a test instance?

Issue #594, opened 2022-12-19
E2EE instability


Half-Shot:

jaller94:
Filed [E2EE: Failed to set up crypto on intent (IO error: could not acquire lock)](https://github.com/matrix-org/matrix-hookshot/issues/609)

Half-Shot:
#714 should help us massively with this.

Half-Shot:
And fixed!

thoraj:
Does this mean that the comment in Hookshot about E2EE instability could be removed? (https://matrix-org.github.io/matrix-hookshot/latest/advanced/encryption.html)  

Or are there more work remaining to get E2EE in Hookshot production ready?

ownerer:
Bump, wondering the same as @thoraj ?

tswfi:
Kinda interested also :)

olmari:
@Half-Shot So, which one is it, fully supported or still highly experimental like https://matrix-org.github.io/matrix-hookshot/latest/advanced/encryption.html still says? How about the current state of needed expoerimentals on synapse side?

ownerer:
I created a new follow-up issue in an attempt to get more visibility (seeing as this issue is closed) and get some clarity on these questions.
See #845

PR #610, opened 2023-01-05
Mark encryption as experimental

Half-Shot:
As discussed in #594, the bridge currently isn't stable with encryption enabled. Rather than ripping all the code out, we're going to mark it as experimental while it's still being worked on.

### DIFF
--- a/config.sample.yml
+++ b/config.sample.yml
@@ -113,11 +113,6 @@ queue:
   monolithic: true
   port: 6379
   host: localhost
-encryption:
-  # (Optional) Configuration for encryption support in the bridge.
-  # If omitted, encryption support will be disabled.
-  #
-  null
 logging:
   # (Optional) Logging settings. You can have a severity debug,info,warn,error
   #
--- a/config.sample.yml
+++ b/config.sample.yml
@@ -117,7 +117,7 @@ encryption:
   # (Optional) Configuration for encryption support in the bridge.
   # If omitted, encryption support will be disabled.
   #
-  storagePath: ./data/encryption
+  null
 logging:
   # (Optional) Logging settings. You can have a severity debug,info,warn,error
   #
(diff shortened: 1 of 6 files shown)


config.sample.yml
```diff
@@ -117,7 +117,7 @@ encryption:
   # (Optional) Configuration for encryption support in the bridge.
   # If omitted, encryption support will be disabled.
   #
-  storagePath: ./data/encryption
+  null
```
AndrewFerr, line 120: How about commenting out the whole section?  I can't remember if the code chokes on having an `encryption` object with no properties, vs an undefined `encryption` object altogether

config.sample.yml
```diff
@@ -117,7 +117,7 @@ encryption:
   # (Optional) Configuration for encryption support in the bridge.
```
AndrewFerr, line 117: Should the option be named to `experimental_encryption` here, or are we leaving it as-is to add even more friction to enabling it?

AndrewFerr: And nitpick, all other config options use camelCase

config.sample.yml
Half-Shot: The null was a bit of a mistake. It was meant to leave this out of the config. We don't yet have support for generating commented out sections (though it's something I want to figure out eventually)

src/Config/Config.ts
```diff
@@ -448,7 +448,9 @@ export class BridgeConfig {
     @configKey("Basic homeserver configuration")
     public readonly bridge: BridgeConfigBridge;
     @configKey(`Configuration for encryption support in the bridge.
- If omitted, encryption support will be disabled.`, true)
+ If omitted, encryption support will be disabled.
+ This feature is HIGHLY EXPERIMENTAL AND SUBJECT TO CHANGE.
+ For more details, see https://github.com/matrix-org/matrix-hookshot/issues/594.`, true)
```
AndrewFerr, line 453: I think this never gets put anywhere since the `generate-default-config` package script is broken (and always has been IIRC), but the same comments are elsewhere, so it's fine.

AndrewFerr (approved):

PR #614, opened 2023-01-06
Decrease connection creation concurrency when starting up

### LINKED ISSUE #608: Network failures on startup can lead to missing connections
Caused in part by #607 

We should probably retry these connections when things fail.

Half-Shot:
Fixes #608 

This isn't really the best possible fix, but I think actually going through and eliminating all the possible ways we can fail a HTTP request on startup to be a significant effort. To avoid spinning too much time away on that, this change:

- Stops us from trying to fetch all state at once on startup at once concurrently, upsetting the homeserver. Instead, load 2 at a time. In the future we can increase this.
- Avoids `getRoomState` failures (which are probably the biggest explosion point) by retrying them on failure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "mime": "^3.0.0",
     "node-emoji": "^1.11.0",
     "nyc": "^15.1.0",
+    "p-queue": "^6.6.2",
     "prom-client": "^14.0.1",
     "reflect-metadata": "^0.1.13",
     "rss-parser": "^3.12.0",
--- a/src/PromiseUtil.ts
+++ b/src/PromiseUtil.ts
@@ -1,3 +1,4 @@
+import { StatusCodes } from "http-status-codes";
 import { Logger } from "matrix-appservice-bridge";
 import { MatrixError } from "matrix-bot-sdk";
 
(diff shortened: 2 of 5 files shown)


src/PromiseUtil.ts
```diff
@@ -1,9 +1,17 @@
 import { Logger } from "matrix-appservice-bridge";
+import { MatrixError } from "matrix-bot-sdk";
 
 const SLEEP_TIME_MS = 250;
 const DEFAULT_RETRY = () => true;
 const log = new Logger("PromiseUtil");
 
+export function retryMatrixErrorFilter(err: unknown) {
+    if (err instanceof MatrixError && err.statusCode >= 400 && err.statusCode <= 499) {
```
jaller94, line 9: It will retry, if the error isn't a MatrixError. Is that what we want?
I thought this is secifically for retrying on (temporary) server errors and rate limiting.

Are gateway timeouts MatrixErrors?

jaller94 (approved):

package.json
```diff
@@ -60,6 +60,7 @@
     "mime": "^3.0.0",
     "node-emoji": "^1.11.0",
     "nyc": "^15.1.0",
+    "p-queue": "^6.6.2",
```
jaller94, line 63: Did yarn.lock not get updated and CI is not complaining?? 🧐

Half-Shot: It was already a dependency of matrix-appservice-bridge, a free one if you will :)

src/PromiseUtil.ts
Half-Shot: The intention is to retry non-fatal errors, i.e. we will retry 5XX errors because more often than not it's the server blowing up. Anything that's not a MatrixError is also considered temporary (we err on the side of caution).

PR #617, opened 2023-01-09
Check the GitLab instance name when looking for existing connections

Half-Shot:
We're checking to see if there are multiple connections with the same `path` in the room, but they might be on different instances!


jaller94 (approved):
Wow. so simple. 🤯
Looks good to me.

PR #620, opened 2023-01-11
Add webhook name and room to admin info

### LINKED ISSUE #619: More info in the „You have bridged a webhook“ message
Use cases are for example looking up URLs for webhooks or to know when and where I configured them.

I would suggest changing

```
You have bridged a webhook. Please configure your webhook source to use https://example.com/
```

to

```
You have bridged the webhook „hook name“ in #room:example.com.
Please configure your webhook source to use https://example.com/
```

weeman1337:
Now in admin webhook info:
- Name of the hook
- Link to room (used a matrix.to link, because that seems to be working)
- URL on own line for better copying

image

closes #619

### DIFF
--- a/src/Connections/SetupConnection.ts
+++ b/src/Connections/SetupConnection.ts
@@ -216,10 +216,11 @@ export class SetupConnection extends CommandConnection {
         this.pushConnections(c.connection);
         const url = new URL(c.connection.hookId, this.config.generic.parsedUrlPrefix);
         const adminRoom = await this.getOrCreateAdminRoom(userId);
+        const safeRoomId = encodeURIComponent(this.roomId);
         await adminRoom.sendNotice(
-            `You have bridged the webhook "${name}" in https://matrix.to/#/${this.roomId} .\n` +
-                // Line break before and no full stop after URL is intentional.
-                // This makes copying and pasting the URL much easier.
+            `You have bridged the webhook "${name}" in https://matrix.to/#/${safeRoomId} .\n` +
+            // Line break before and no full stop after URL is intentional.
+            // This makes copying and pasting the URL much easier.
             `Please configure your webhook source to use\n${url}`
         );
         return this.as.botClient.sendNotice(this.roomId, `Room configured to bridge webhooks. See admin room for secret url.`);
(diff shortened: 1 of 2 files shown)


src/Connections/SetupConnection.ts
```diff
@@ -216,7 +216,12 @@ export class SetupConnection extends CommandConnection {
         this.pushConnections(c.connection);
         const url = new URL(c.connection.hookId, this.config.generic.parsedUrlPrefix);
         const adminRoom = await this.getOrCreateAdminRoom(userId);
-        await adminRoom.sendNotice(`You have bridged a webhook. Please configure your webhook source to use ${url}.`);
+        await adminRoom.sendNotice(
+            `You have bridged the webhook "${name}" in https://matrix.to/#/${this.roomId} .\n` +
```
jaller94, line 220: While an official Synapse will generate safe room IDs, the spec doesn't seem to limit what characters might be included.

matrix.to can deal with URL-encoded room IDs and room aliases.

```suggestion
            `You have bridged the webhook "${name}" in https://matrix.to/#/${encodeURIComponent(this.roomId)} .\n` +
```

Alternatively, if you think it looks too weird that ":" gets encoded:
```suggestion
            `You have bridged the webhook "${name}" in https://matrix.to/#/${encodeURIComponent(this.roomId).replace(/%3A/g, ":")} .\n` +
```

src/Connections/SetupConnection.ts
```diff
@@ -216,7 +216,12 @@ export class SetupConnection extends CommandConnection {
...
+                // Line break before and no full stop after URL is intentional.
+                // This makes copying and pasting the URL much easier.
```
jaller94, line 222: You may have these 8 spaces back to put them somewhere else. :yum: 
```suggestion
            // Line break before and no full stop after URL is intentional.
            // This makes copying and pasting the URL much easier.
```

jaller94 (approved):
Looks good to me. Accepting the suggestions would be highly appreciated.

src/Connections/SetupConnection.ts
Half-Shot: Former looks fine to me :)

src/Connections/SetupConnection.ts
weeman1337: Since we started using Prettier in other repos, I have become really lazy :upside_down_face:

PR #624, opened 2023-01-13
New project picker component for room configs

Half-Shot:
This refactors the dropdown that lets you pick a project/repo in the room config widget into something that:

 - Gives you more context (name,avatar and description) about what you're selecting.
 - Allows you to see more easily what you have selected.
 - Avoids bugs present with the autosuggestions feature on standard HTML inputs like clipping through the iframe, not updating when the input changes, and generally feeling unresponsive.

This change affects GitHub, GitLab and Jira.

NOW WITH A DEMO:

https://user-images.githubusercontent.com/2072976/213426884-36b9cd2b-5591-4f73-a4f5-0b92b94ab2e4.mp4


web/components/elements/DropdownSearch.module.scss
```diff
@@ -0,0 +1,39 @@
+.dropdownItem {
+    padding: 0.25em 0.5em;
+    border-radius: 8px;
+}
+
+.dropdownItem:hover {
+    cursor: pointer;
+    background-color: rgba(54, 54, 54, 0.138);
+}
```
jaller94, line 9: Let me tell you about the magic and wonders of nested SCSS.
Also, `cursor` only affects the cursor on hover anyway.

```suggestion
    cursor: pointer;

    &:hover {
        background-color: rgba(54, 54, 54, 0.138);
    }
}
```

web/components/elements/DropdownSearch.tsx
```diff
@@ -0,0 +1,95 @@
...
+            console.log("searching for ", searchTerm);
+            searchFn(searchTerm, searchProps).then(result => {
+                setResults(result);
+            }).catch(err => {
+                onError?.(err);
+            })
+        }, DEBOUNCE_TIMEOUT_MS) as unknown as number);
+        return () => {
+            clearTimeout(debounceTimer);
+        }
```
jaller94, line 68: Why is `debounceTimer` a state?
I think this will reduce re-renders and prevent potential memory leaks (when the timer gets re-set in quick concession.

```suggestion
        const debounceTimer = setTimeout(() => {
            console.log("searching for ", searchTerm);
            searchFn(searchTerm, searchProps).then(result => {
                setResults(result);
            }).catch(err => {
                onError?.(err);
            })
        }, DEBOUNCE_TIMEOUT_MS);
        return () => {
            clearTimeout(debounceTimer);
        }
```

web/components/elements/DropdownSearch.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+        // Browser types
+        setDebounceTimer(setTimeout(() => {
+            console.log("searching for ", searchTerm);
+            searchFn(searchTerm, searchProps).then(result => {
+                setResults(result);
+            }).catch(err => {
+                onError?.(err);
+            })
```
jaller94, line 64: Minor nitpick but likely a rare edge that users know to deal with: Once the search has been started, a slow network fetch will override a fast one.
It will also set the results if `onItemPicked` has set the results to `[]` or `searchProps` changed in the meantime.

web/components/elements/DropdownSearch.tsx
```diff
@@ -0,0 +1,95 @@
...
+        setResults([]);
+    }, [onChange]);
+
+
+    return <>
+        <input type="search" placeholder={placeholder} onChange={onSearchInputChange} value={searchTerm} />
+        {selectedItem && <DropdownItem {...selectedItem} />}
+        <ul>
+            {
+                results.map(item => <DropdownItem key={item.value} {...item} onPicked={ () => onItemPicked(item) } />)
```
jaller94, line 91: Not using `useCallback` to memorize a function handler might cause performance issues on low end devices.
Changes to the DOM are expensive.

web/components/roomConfig/GitlabRepoConfig.tsx
```diff
@@ -6,94 +6,110 @@ import { GitLabRepoConnectionState, GitLabRepoResponseItem, GitLabTargetFilter,
...
+            return;
+        }
+        if (!currentInstance) {
+            throw Error('Should never pick a project without an instance');
+        }
+        onPicked({
+            instance: currentInstance,
+            path: value,
+        })
+    }, [currentInstance, onPicked]);
```
jaller94, line 95:
```suggestion
        onPicked(state => {
            if (!state.instance) {
                throw Error('Should never pick a project without an instance');
            }
            return {
                ...state,
                path: value,
            }
        });
    }, [onPicked]);
```

jaller94 (approved):
Couldn't find any big issues. Just super minor performance issues for super low end devices.

Half-Shot:
(It's still a draft FYI, I still need to work on this a bit :) )

src/Connections/GithubRepo.ts
```diff
@@ -1258,6 +1261,48 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
...
+                break;
+            }
+            page++;
+            installationsCount += data.repositories.length;
+            totalCount = data.total_count;
+            // Find any repos that were in our search results. If a search term isn't defined, just return it.
+            foundRepos.push(...data.repositories.filter((installRepo) => searchRepos?.includes(installRepo.full_name) ?? true));
+        } while (
+            installationsCount < totalCount &&
+            foundRepos.length < (searchRepos?.length ?? MAX_RETURNED_TARGETS)
```
jaller94, line 1302: `foundRepos` may contain more items than `MAX_RETURNED_TARGETS` which makes the variable name misleading.

src/Connections/GitlabRepo.ts
```diff
@@ -291,7 +295,39 @@ export class GitLabRepoConnection extends CommandConnection<GitLabRepoConnection
...
+        } catch (ex) {
+            if (axios.isAxiosError(ex)) {
+                if (ex.response?.status === 401) {
+                    // 401 means that the project is Private and GitLab haven't fixed
+                    // the auth issues, just ignore this one.
+                    // https://gitlab.com/gitlab-org/gitlab/-/issues/25498
+                    return null;
+                }
+            }
+            log.warn(`Could not transform data from ${avatarUrl} into base64`, ex);
```
jaller94, line 325: You only tested that it's not an HTTP 404 error.
This could still be an HTTP 500 or other run time error.

Also, I don't think there is any Buffer which cannot be transformed to base64.

web/components/elements/ConnectionSearch.tsx
```diff
@@ -0,0 +1,127 @@
...
+}: IProps) {
+    const [currentInstance, setCurrentInstance] = useState<string>("");
+    const [instances, setInstances] = useState<Instance[]|null>(null);
+    const [searchError, setSearchError] = useState<string|null>(null);
+    const [exampleProjectName, setExampleProjectName] = useState<string>("Loading...");
+
+    useEffect(() => {
+        getInstances().then(res => {
+            setInstances(res);
+            setCurrentInstance(res[0]?.name ?? null);
```
jaller94, line 43: This state got defined as a string with an empty string being the default value.
Good morning, TypeScript linter. :wave: 

```suggestion
            setCurrentInstance(res[0]?.name ?? '');
```

src/Connections/GithubRepo.ts
```diff
@@ -1258,6 +1261,48 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
...
         }
     }
+    
+    public static async searchInstallationForRepos(octokit: Octokit, orgName: string, installationId: number, searchTerms?: string) {
+        // First, do a search on GitHub for repos. This will use the user's context so it will find user repos.
+        let searchRepos: string[]|null = null;
+        if (searchTerms) {
+            const terms = encodeURIComponent(searchTerms);
+            const searchResultsData = (await octokit.search.repos({
+                q: `${terms} org:${orgName} `,
```
jaller94, line 1271: Could any user abuse `term` to query terms in other orgs with Hookshot's GitHub token?
e.g. `term = "hl3 org:valve"`

No, because it's queried by the user's webbrowser, right?

Half-Shot: The users access token yeah.

web/components/elements/DropdownSearch.module.scss
Half-Shot: > Also, cursor only affects the cursor on hover anyway.

I am so dumb :D

web/components/elements/DropdownSearch.tsx
Half-Shot: Yeah, that makes a ton of sense! Updating.

web/components/elements/DropdownSearch.tsx
Half-Shot: Yeah :( This is one of those pain in the butt cases where we want to create a function for each item.

src/Connections/GithubRepo.ts
Half-Shot: I think this is okay. The initial search can return at most `MAX_RETURNED_TARGETS`, and then our filter code can return only items in that search set. There shouldn't be dupes, unless GitHub's API returns dupes.

src/Connections/GitlabRepo.ts
Half-Shot: It's okay for this function to fail for any reason, we just want to log if the error was unexpected. However, let's improve the error reporting.

Half-Shot:
Design did a brief review of this in DMs, basically needs some polish (to be done when we go back and look at hookshot with our design systems hats on), but the only major change was the "Your connected projects" line being removed when you have no projects.

PR #628, opened 2023-01-15
Fix state update validation

### LINKED ISSUE #596: `transformationFunction` updates not applied
**Steps to reproduce the problem**

- Room with a configured webhook and `transformationFunction` set
- Update `transformationFunction`
- Trigger the webhook

**Expected:**

New transformation function applied

**What actually happens:**

Old transformation function applied

image

image

weeman1337:
One single char can make a big difference :wink: 

closes #596

### DIFF
--- a/changelog.d/628.bugfix
+++ b/changelog.d/628.bugfix
@@ -1,0 +1,1 @@
+Hookshot now handles `uk.half-shot.matrix-hookshot.generic.hook` state event updates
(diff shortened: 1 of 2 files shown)


Dumme34:
😂😂😂👍

Half-Shot (approved):
GAH, so that was what was causing the pain. Thank you.

Issue #609, opened 2023-01-05
Failed to set up crypto on intent (IO error: could not acquire lock)


jaller94:
Hookshot was already running for 2+ hours when it crashed with this message.

## Crash example 1
```
INFO 16:56:44:962 [ConnectionManager] Looking to provision connection for !MBKReTGbuldfQteUbW:localhost uk.half-shot.matrix-hookshot.feed for @jaller94:localhost with data {"url":"https://chrpaul.de/feed.xml","label":"Chr Paul"}
INFO 16:56:45:120 [FeedConnection] Connection 0e343504299411063c7324cc5166baa8 created for !MBKReTGbuldfQteUbW:localhost, {"url":"https://chrpaul.de/feed.xml","label":"Chr Paul"}
ERROR 16:56:45:246 [Appservice] Failed to set up crypto on intent @hookshot:localhost [Error: IO error: could not acquire lock on "/home/jaller94/Git/matrix-hookshot/data/encryption/2d03b2ce09b8e551150e45d1418338aa0f43c531f79e8a34a1758bcbbbc850d0/matrix-sdk-crypto/db": Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }] {
  code: 'GenericFailure'
}
Error: IO error: could not acquire lock on "/home/jaller94/Git/matrix-hookshot/data/encryption/2d03b2ce09b8e551150e45d1418338aa0f43c531f79e8a34a1758bcbbbc850d0/matrix-sdk-crypto/db": Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }
error Command failed with exit code 1.
```

## Crash example 2
```
INFO 10:11:30:044 [Appservice] Processing transaction 7820
INFO 10:11:30:044 [Appservice] Processing ephemeral event of type m.presence
INFO 10:11:30:045 [Appservice] 127.0.0.1 - - [03/Jan/2023:09:11:30 +0000] "PUT /transactions/7820?access_token=%3Credacted%3E HTTP/1.1" 200 2 "-" "Synapse/1.74.0"

[...]

PR #633, opened 2023-01-18
Set LRU values to zero to prevent cleanup when using encryption

Half-Shot:
Requires https://github.com/vector-im/matrix-bot-sdk/pull/6

Should fix https://github.com/matrix-org/matrix-hookshot/issues/609, although other process crashing bugs may remain.

### DIFF
--- a/src/appservice.ts
+++ b/src/appservice.ts
@@ -38,6 +38,10 @@ export function getAppservice(config: BridgeConfig, registration: IAppserviceReg
         },
         storage: storage,
         intentOptions: {
+            // If encryption support is enabled, we cannot expire Intent objects or we risk
+            // a resource contention on the Sled DB.
+            maxCached: config.encryption && 0,
+            maxAgeMs: config.encryption && 0,
             encryption: !!config.encryption,
         },
         cryptoStorage: cryptoStorage,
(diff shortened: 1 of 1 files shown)


src/appservice.ts
```diff
@@ -38,6 +38,10 @@ export function getAppservice(config: BridgeConfig, registration: IAppserviceReg
         },
         storage: storage,
         intentOptions: {
+            // If encryption support is enabled, we cannot expire Intent objects or we risk
+            // a resource contention on the Sled DB.
+            maxCached: config.encryption && 0,
+            maxAgeMs: config.encryption && 0,
```
jaller94, line 44: Better use something like `config.encryption ? foo : 0` to limit what falsey values are passed to the Intent.
```
> true && 0
0
> false && 0
false
> null && 0
null
> undefined && 0
undefined
```

jaller94 (approved):
Yes, but please limit possible types.

src/appservice.ts
Half-Shot: Don't we trust `config.encryption` to be object|undefined though, as that's strongly typed in the config.

Half-Shot:
We're not going to do this, in favour of replacing Sled.

Issue #641, opened 2023-02-12
Don't provision service bots when the respective service is disabled


HarHarLinks:
Context: https://matrix.to/#/!TlZdPIYrhwNvXlBiEk:half-shot.uk/$AdcR43mfQlQq2p97UOgmIZTfUlmBVNbI7CtTQZ6_-74?via=half-shot.uk&via=matrix.org&via=element.io

In the sample config, feeds are disabled but a feeds service bot is enabled. This leads to errors when the service bot isn't namespaced in the registration.yaml

Half-Shot:
It should probably be a startup failure to configure service bots for disabled services :thinking:

PR #643, opened 2023-02-12
Register namespace for sample serviceBot in registration.sample.yml

HarHarLinks:

Context: https://matrix.to/#/!TlZdPIYrhwNvXlBiEk:half-shot.uk/$AdcR43mfQlQq2p97UOgmIZTfUlmBVNbI7CtTQZ6_-74?via=half-shot.uk&via=matrix.org&via=element.io

Sample config
- disables feeds
- enables feeds serviceBot
- does not register feeds serviceBot namespace

This change fixes the latter, thus providing an example for anybody using servicebots.

[The spec](https://spec.matrix.org/unstable/application-service-api/#registration) isn't totally clear that this is needed, but it appears to be this way with synapse. Should the spec be updated to state this requirement more clearly?

Related: https://github.com/matrix-org/matrix-hookshot/issues/641

### DIFF
--- a/changelog.d/643.doc
+++ b/changelog.d/643.doc
@@ -1,0 +1,1 @@
+Update docs and sample config for serviceBots. Thanks to @HarHarLinks.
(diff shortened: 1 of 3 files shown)


Half-Shot (approved):
Looks great to me.

PR #647, opened 2023-02-17
Add a go-neb migrator

tadzik:
Deals only with feeds so far, but it should be easier to add more things to it now.

[migrator-2023-02-17_14.47.26.webm](https://user-images.githubusercontent.com/86378/219674697-212a91c4-3433-44b4-bae0-a3e9cc3088ba.webm)

The migrator converts go-neb service configs to hookshot-compatible connection states. Frontend grabs them, allows for adjustments, and saves them natively in hookshot (client-side, as if we handcrafted them).

The original "connections" are *not* removed from go-neb.

Some data gets lost - notably features that go-neb supported but we don't, such as:

```
        "must_include": {
          "author": null,
          "title": null,
          "description": null
        },
        "must_not_include": {
          "author": null,
          "title": null,
          "description": null
        },
```

for feeds.

We may want to store them for later, but that'd require soft-commiting to supporting them later, and in this particular format.

### DIFF
--- a/src/Config/Config.ts
+++ b/src/Config/Config.ts
@@ -8,7 +8,7 @@ import { GitHubRepoConnectionOptions } from "../Connections/GithubRepo";
 import { BridgeConfigActorPermission, BridgePermissions } from "../libRs";
 import { ConfigError } from "../errors";
 import { ApiError, ErrCode } from "../api";
-import { GITHUB_CLOUD_URL } from "../Github/GithubInstance";
+import { GithubInstance, GITHUB_CLOUD_URL } from "../Github/GithubInstance";
 import { Logger } from "matrix-appservice-bridge";
 
 const log = new Logger("Config");
@@ -95,10 +95,10 @@ export class BridgeConfigGitHub {
         this.baseUrl = yaml.enterpriseUrl ? new URL(yaml.enterpriseUrl) : GITHUB_CLOUD_URL;
     }
 
-    @hideKey()
-    public get publicConfig() {
+    public publicConfig(githubInstance?: GithubInstance) {
         return {
             userIdPrefix: this.userIdPrefix,
+            newInstallationUrl: githubInstance?.newInstallationUrl?.toString(),
         }
     }
 }
@@ -716,7 +716,7 @@ For more details, see https://github.com/matrix-org/matrix-hookshot/issues/594.
                 config = this.generic?.publicConfig;
                 break;
             case "github":
-                config = this.github?.publicConfig;
+                config = this.github?.publicConfig();
                 break;
             case "gitlab":
                 config = this.gitlab?.publicConfig;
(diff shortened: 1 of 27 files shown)


src/Widgets/BridgeWidgetApi.ts
```diff
@@ -54,9 +57,36 @@ export class BridgeWidgetApi {
         this.api.addRoute("patch", '/v1/:roomId/connections/:connectionId', wrapHandler(this.updateConnection));
         this.api.addRoute("delete", '/v1/:roomId/connections/:connectionId', wrapHandler(this.deleteConnection));
         this.api.addRoute("get", '/v1/targets/:type', wrapHandler(this.getConnectionTargets));
+
+        if (this.config.goNebMigrator) {
+            this.goNebMigrator = new GoNebMigrator(
```
justinbot, line 62: I don't see this class, maybe forgot to commit it?

src/Config/Config.ts
```diff
@@ -431,6 +431,11 @@ export interface BridgeConfigMetrics {
     port?: number;
 }
 
+export interface BridgeConfigGoNebMigrator {
+    apiUrl: string;
+    serviceIds: string[];
```
justinbot, line 436: These service IDs are unfortunately not static, but generated each time a go-neb bot configuration is created by a user (with the idea being you generate the same ID later to retrieve the configuration):

```ts
`${type}/${strictEncodeURIComponent(userId)}/${strictEncodeURIComponent(roomId)}`;
```

Where `type` is:
* `github` for Github (annoyingly **not** `github-webhook` like the service Type is named)
* `rssbot` for Feeds.

and `userId` is the user who originally added the Go-NEB bot.

and `strictEncodeURIComponent` is RFC3986 compliant like this:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#encoding_for_rfc3986

web/components/roomConfig/RoomConfig.tsx
```diff
@@ -149,5 +171,16 @@ export const RoomConfig = function<SConfig, ConnectionType extends GetConnection
                 </ListItem>)
             }
         </section>}
+        { toMigrate.length > 0 && <section>
+            <h2> Migrate connections </h2>
+            { serviceConfig && toMigrate.map(c => <ListItem key={JSON.stringify(c)} text={listItemName(c)}>
```
justinbot, line 176: Minor, but I think each item `toMigrate` has an `id` we can use for the key instead?

justinbot (changes requested):
Some changes needed but looking good so far.

I think some way to dismiss/delete migrations will improve the UX significantly because having these candidates never go away seems annoying.
The state event approach seems okay, but is there a reason we shouldn't just delete them from go-neb?

justinbot (approved):
A few more suggestions but otherwise looks good

PR #655, opened 2023-03-08
Refactor hookshot to include permission grant system 

Half-Shot:
This change is huge but should hopefully bring about better support for self service public hookshots. This changes the code so that on connection creation, we record a grant in hookshot so that when we come to rehydrate the state later on restart, we can verify access of the connection's resource.

### Review guide

There are several components to this system:

Firstly, the `GrantChecker` class. It basically works by reading and storing grants in account data for hookshot, ensuring that we have recorded a given roomId has access to a thing. A thing is defined by `ConnnectionId`, which is how we determine the resource to be saved for each different type of service.

Since we've not been recording the grants up until now, we need *some* way to infer whether existing connections are valid. This is where the fallback system kicks in. For the fallbacks, we allow connections to either check an external service (GitHub, GitLab) or just look at the bridge permissions in the config. If the fallback determines a grant should be given, one is stored.

And then finally, each connection has been updated to check, store and remove grants now. Connection manger will call `ensureGrant` prior to committing a connection to memory to ensure it's valid, but it's also imperative that connections check on each state update for this validity.


src/grants/GrantCheck.ts
```diff
@@ -0,0 +1,174 @@
...
+    org: string;
+    repo: string;
+}
+
+interface GitLabGrantConnectionId{
+    instance: string;
+    path: string;
+}
+
+type ConnectionId = NonNullable<string|object>;
```
justinbot, line 33: I think `NonNullable` is redundant here?

src/grants/GrantCheck.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /**
+     * Check the permissions on GitHub, in case of a missing grant.
+     * @param as The appservice.
+     * @param config The bridge config.
+     * @param service The service name e.g. "github"
+     * @returns A GrantChecker
+     */
+    static withGitHubFallback(as: Appservice, github: GithubInstance, tokenStore: UserTokenStore) {
```
justinbot, line 71: Could each connection pass in a fallback function rather than centralizing them all in here?

src/Connections/GithubUserSpace.ts
```diff
@@ -137,12 +143,21 @@ export class GitHubUserSpace extends BaseConnection implements IConnection {
...
-    constructor(public readonly space: Space,
+    private readonly grantChecker: GrantChecker;
+
+    constructor(as: Appservice,
+        config: BridgeConfig,
+        public readonly space: Space,
         private state: GitHubUserSpaceConnectionState,
         stateKey: string) {
             super(space.roomId, stateKey, GitHubUserSpace.CanonicalEventType);
+            this.grantChecker = GrantChecker.withConfigFallback(as, config, "github");
```
justinbot, line 154: Could we have one `GrantChecker` instance that gets passed to each connection, instead of having to instantiate it everywhere with the same `as` and `config`?

src/grants/GrantCheck.ts
```diff
@@ -0,0 +1,174 @@
...
+        });
+    }
+
+    private static stringifyConnectionId<cId = ConnectionId>(connId: cId) {
+        if (typeof connId === "string") {
+            return FormatUtil.hashId(connId.toString());
+        }
+        if (Array.isArray(connId)) {
+            throw Error('Array types are invalid');
+        }
```
justinbot, line 122: Doesn't look like `cId` could ever be an array?

justinbot (commented):
First look, broadly makes sense.
A bit concerned about (1) how much boilerplate each connection seems to require and (2) adding more connection-specific code outside of connections. Hopefully both can be avoided with a bit of refactoring.

src/grants/GrantCheck.ts
```diff
@@ -0,0 +1,176 @@
+import { Logger } from "matrix-appservice-bridge";
+import { Appservice, Intent, MatrixError } from "matrix-bot-sdk";
+import { BridgeConfig, BridgeConfigGitLab, BridgePermissionLevel } from "../Config/Config";
+import { GitHubRepoConnection, GitLabRepoConnection } from "../Connections";
```
Half-Shot, line 4: Dropping into a thread (@justinbot):

> (2) adding more connection-specific code outside of connections. Hopefully both can be avoided with a bit of refactoring

This was a hard choice. The problem I was facing is that I didn't want to repeat that logic N times for every GitHub connection type, so instead I pasted it inside the grant code. We could probably stop cheating and create subclasses of GrantChecker and not store them alongside the grant checker itself.

src/grants/GrantCheck.ts
```diff
@@ -0,0 +1,176 @@
+import { Logger } from "matrix-appservice-bridge";
```
Half-Shot, line 1: Dropping into a thread (@justinbot):

>  (1) how much boilerplate each connection seems to require 

So from my side I've tried to reduce the boilerplate as much as I can, but each connection sort of needs to do different things in subtle but annoying ways.  I originally tried to make this part of the ConnectionManager although it was difficult due to each requirement, so it ended up per connection with as much logic as I could manage abstracted into the grantChecker class. 

Of course, I've been looking at this for so long there might be another way.

src/Connections/GithubDiscussion.ts
```diff
@@ -94,9 +94,19 @@ export class GitHubDiscussionConnection extends BaseConnection implements IConne
...
         await intent.ensureJoined(roomId);
+
         return new GitHubDiscussionConnection(roomId, as, intent, state, '', tokenStore, commentProcessor, messageClient, config);
     }
 
+    private static grantKey(state: GitHubDiscussionConnectionState) {
+        return `${this.CanonicalEventType}/${state.owner}/${state.repo}`;
+    }
+
+    private readonly sentEvents = new Set<string>(); //TODO: Set some reasonable limits
```
justinbot, line 105: TODO needs addressing?

src/grants/GrantCheck.ts
justinbot: This makes sense, they're each a bit different. It's not the worst boilerplate, just something I've noticed we have more and more of.

src/grants/GrantCheck.ts
justinbot: I was thinking we could take a more functional approach, from my other comment:
> Could each connection pass in a fallback function rather than centralizing them all in here?

But I'm fine to leave this for a later refactor rather than changing things around now.

justinbot (approved):

src/Connections/GithubDiscussion.ts
Half-Shot: I think this was already here but...we can address it anyway :)

PR #656, opened 2023-03-09
Handle Scalar-style go-neb service IDs

tadzik:
Fixing https://github.com/matrix-org/matrix-hookshot/pull/647#discussion_r1120766260

### DIFF
--- a/src/Widgets/GoNebMigrator.ts
+++ b/src/Widgets/GoNebMigrator.ts
@@ -125,10 +125,10 @@ export class GoNebMigrator {
                         }
                     }
                     break;
-                };
+                }
                 default: {
                     log.warn(`Unrecognized go-neb service type (${obj.Type}), skipping`);
-                };
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/Widgets/GoNebMigrator.ts
```diff
@@ -77,17 +84,34 @@ export class GoNebMigrator {
     }
 
     public async getConnectionsForRoom(roomId: string, userId: string): Promise<MigratedConnections> {
-        const feeds: FeedConnectionState[] = [];
-        const github: GitHubRepoConnectionState[] = [];
+        const feeds: MigratedFeed[] = [];
+        const github: MigratedGithub[] = [];
 
-        for (const id of this.serviceIds) {
+        const services = [
```
justinbot, line 90: I think `serviceIds` would be more accurate

justinbot (changes requested):

justinbot (approved):

PR #657, opened 2023-03-10
Fix widget config detection

weeman1337:
Explicitly check for `undefined` when testing `roomSetupWidget`.

From the docs: „This can be enabled by setting roomSetupWidget to an object. “

If you set this to an object in the config:

```
widgets:
  publicUrl: http://localhost:9000/widgetapi/v1/static
  branding:
    widgetTitle: Hookshot Konfig
  roomSetupWidget:
```

The value of the parsed `roomSetupWidget` will be null → check fails


jaller94 (approved):
Looks good to me.

PR #659, opened 2023-03-13
Fix service bots not rejecting invites on incoming DMs

Half-Shot:
Synapse no longer lets you kick yourself from a room with a reason, so instead we'll do the spec compliant thing of sending a leave with a reason. 

https://github.com/turt2live/matrix-bot-sdk/pull/301 is the upstream change we could rely on if it turned up in time.

### DIFF
--- a/src/Bridge.ts
+++ b/src/Bridge.ts
@@ -789,18 +789,24 @@ export class Bridge {
             // We got an invite but it's not a configured bot user, must be for a ghost user
             log.debug(`Rejecting invite to room ${roomId} for ghost user ${invitedUserId}`);
             const client = this.as.getIntentForUserId(invitedUserId).underlyingClient;
-            return client.kickUser(invitedUserId, roomId, "Bridge does not support DMing ghosts");
+            return client.doRequest("POST", "/_matrix/client/v3/rooms/" + encodeURIComponent(roomId) + "/leave", null, {
+                reason: "Bridge does not support DMing ghosts"
+            });
         }
 
         // Don't accept invites from people who can't do anything
         if (!this.config.checkPermissionAny(event.sender, BridgePermissionLevel.login)) {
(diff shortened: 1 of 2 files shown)


src/Bridge.ts
```diff
@@ -789,18 +789,24 @@ export class Bridge {
...
+            return botUser.intent.underlyingClient.doRequest("POST", "/_matrix/client/v3/rooms/" + encodeURIComponent(roomId) + "/leave", null, {
+                reason: "You do not have permission to invite this bot."
+            });
         }
 
         if (event.content.is_direct && botUser.userId !== this.as.botUserId) {
             // Service bots do not support direct messages (admin rooms)
             log.debug(`Rejecting direct message (admin room) invite to room ${roomId} for service bot ${botUser.userId}`);
-            return botUser.intent.underlyingClient.kickUser(botUser.userId, roomId, "This bot does not support admin rooms.");
+            return botUser.intent.underlyingClient.doRequest("POST", "/_matrix/client/v3/rooms/" + encodeURIComponent(roomId) + "/leave", null, {
```
justinbot, line 807: I'm a bit surprised there is no `leave` method in the `matrix-bot-sdk` client 🙁

justinbot (approved):

src/Bridge.ts
```diff
@@ -789,18 +789,24 @@ export class Bridge {
...
+                reason: "This bot does not support admin rooms."
+            });
```
justinbot, line 809: Would it make sense to actually send this message in the room before leaving?

Element doesn't actually display the leave reason in the UI, although you could argue it should.

Half-Shot: If it doesn't show a reason in the membership event, it's not being terribly spec compliant. I'd argue it should fix it.

PR #667, opened 2023-03-21
Add release script

Half-Shot:
This basically ports https://github.com/matrix-org/matrix-appservice-bridge/pull/452 although I've made it a bit more agnostic to any package, so it should reasonably work if we ported it back.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -44,8 +44,6 @@ echo "Generated tag $TAG"
 
 echo "Pushing to origin"
 git push origin $TAG
-
-# Push to main branch too
 git push
 
 echo "The CI to generate a release is now running. Check https://github.com/$REPO_NAME/releases and publish the release when it's ready."
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -35,16 +35,17 @@ fi
 
 echo "Committing version"
 towncrier build --version $VERSION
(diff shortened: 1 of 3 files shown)


scripts/release.sh
```diff
@@ -0,0 +1,49 @@
...
+
+read -p "Happy with the changelog? <y/N> " prompt
+if [[ $prompt != "y" && $prompt != "Y" && $prompt != "yes" && $prompt != "Yes" ]]
+then
+  rm draft-release.txt
+  exit 0
+fi
+
+echo "Committing version"
+towncrier build --version $VERSION
```
justinbot, line 37: Is `build` a valid command? I've hit this error before running `towncrier`:
```
> towncrier build
Usage: towncrier [OPTIONS]
Try 'towncrier --help' for help.

Error: Got unexpected extra argument (build)
```

But if it works then 🤷

justinbot (approved):
Still wish it was fully automated, but this is a step in the right direction 👍

scripts/release.sh
Half-Shot: Erm, lemme double check.

Half-Shot: My local environment likes it, hmm :/

Half-Shot: I just ran the command on latest towncrier and it was happy, so I think this is fine.

PR #670, opened 2023-03-22
Make the feed parser less strict on HTML data.

Half-Shot:

### DIFF
--- a/src/feeds/FeedReader.ts
+++ b/src/feeds/FeedReader.ts
@@ -197,7 +197,14 @@ export class FeedReader {
                     this.cacheTimes.set(url, { lastModified: res.headers['Last-Modified'] });
                 }
 
-                const feed = await (new Parser()).parseString(res.data);
+                const feed = await (new Parser({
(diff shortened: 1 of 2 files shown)


src/feeds/FeedReader.ts
```diff
@@ -197,7 +197,14 @@ export class FeedReader {
                     this.cacheTimes.set(url, { lastModified: res.headers['Last-Modified'] });
                 }
 
-                const feed = await (new Parser()).parseString(res.data);
+                const feed = await (new Parser({
```
justinbot, line 200: Do we need to instantiate a new parser every time?

justinbot (approved):

src/feeds/FeedReader.ts
Half-Shot: Good question, probably not. I wonder if it's to prevent contamination.

Half-Shot: Looked okay, have refactored.

PR #675, opened 2023-03-23
Don't hide form elements in GitHub component if we're not checking auth

Half-Shot:
This should fix https://github.com/vector-im/element-integration-manager/issues/50


jaller94 (approved):
Looks good to me.

PR #681, opened 2023-03-24
Add feeds failing metric

Half-Shot:
A simple counter that lets us see whether we are having HTTP failures or parsing failures, to track relative success when we modify the parser.


jaller94 (approved):
Looks good to me.

PR #682, opened 2023-03-24
Add a `type` label to metric `hookshot_connections`

Half-Shot:

### DIFF
--- a/src/ConnectionManager.ts
+++ b/src/ConnectionManager.ts
@@ -61,10 +61,14 @@ export class ConnectionManager extends EventEmitter {
             }
             this.connections.push(connection);
             this.emit('new-connection', connection);
+
+            const cd: ConnectionDeclaration = Object.getPrototypeOf(connection).constructor;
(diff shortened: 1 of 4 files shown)


src/ConnectionManager.ts
```diff
@@ -61,11 +61,15 @@ export class ConnectionManager extends EventEmitter {
             }
             this.connections.push(connection);
             this.emit('new-connection', connection);
+
+            const cd: ConnectionDeclaration = Object.getPrototypeOf(connection).constructor;
+            const canonicalEvent = cd?.EventTypes?.[0] ?? 'unknown';
```
jaller94, line 66: > `Object.getPrototypeOf(connection).constructor`

Just because JavaScript allows you to do this, doesn't mean you should.
If there's no other simple way to accomplish this, please add an inline comment of what this is doing.

The following things are uncommon about this (aka. I've never seen them used in a JavaScript project):
* getting the prototype of an object
* assigning the constructor of a class to a new variable
* not executing a constructor, but instead using it like an object variable to get data from the class

PR #683, opened 2023-03-24
Undo XML parser customizations

tadzik:
These did more harm then good, breaking many legit feeds in weird ways.


Half-Shot:
(Accepting change w/o changelog, we're removing a change made by #670 )

Issue #687, opened 2023-03-25
The feed bot keeps repeating some items


subbink:
I had subscribed to [bits of freedom's feed](https://www.bitsoffreedom.nl/feed/) using the migration option of the new feed bot.
This feed keeps repeating some items with every fetch.

I would expect that it would only show new items if they appear.

At first glance it seems that items without a category tag are repeated every time.

[bof-feed.xml.gz](https://github.com/matrix-org/matrix-hookshot/files/11068074/bof-feed.xml.gz) is a copy of the xml that caused an issue.

At the moment I have unsubscribed from the feed to keep other feeds visible.

Please let me know if I can help with testing.

Half-Shot:
Looks like this got done!

subbink:
I can confirm that the feed no longer repeats items.

PR #689, opened 2023-03-27
Make sure we're not treating garbage data in feed item as guids

tadzik:
Some of these would parse as `{ '$': { isPermaLink: 'false' } }` or similar, which would then surprise us very much when assuming that it's a string stored in accountData.

Should fix the problems in https://github.com/matrix-org/matrix-hookshot/issues/687.

### DIFF
--- a/changelog.d/691.bugfix
+++ b/changelog.d/691.bugfix
@@ -1,0 +1,1 @@
+Improve resiliency to invite/join races when Hookshot is added by an integration manager.
--- a/src/Bridge.ts
+++ b/src/Bridge.ts
@@ -818,7 +818,14 @@ export class Bridge {
         }
 
         // Accept the invite
-        await retry(() => botUser.intent.joinRoom(roomId), 5);
+        await retry(async () => {
+            try {
+                await botUser.intent.joinRoom(roomId);
+            } catch (ex) {
+                log.warn(`Failed to join ${roomId}`, ex);
+                throw ex;
+            }
+        }, 5);
         if (event.content.is_direct) {
             await botUser.intent.underlyingClient.setRoomAccountData(
                 BRIDGE_ROOM_TYPE, roomId, {admin_user: event.sender},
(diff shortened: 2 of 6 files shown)


src/feeds/FeedReader.ts
```diff
@@ -246,7 +246,9 @@ export class FeedReader {
                 const newGuids = [];
                 log.debug(`Found ${feed.items.length} entries in ${url}`);
                 for (const item of feed.items) {
-                    const guid = item.guid || item.id || item.link || item.title;
+                    // Find the first guid-like that looks like a string.
+                    // Some feeds have a nasty habit of leading a empty tag there, making us parse it as garbage.
+                    const guid = [item.guid, item.id, item.link, item.title].find(id => typeof id === 'string');
```
jaller94, line 251: Maybe unintended change:
* Empty strings in one of the earlier values now prevent later valid values from being used.

To exclude empty strings early:
```suggestion
                    const guid = [item.guid, item.id, item.link, item.title].find(id => typeof id === 'string' && id);
```

src/feeds/FeedReader.ts
```diff
@@ -246,7 +246,9 @@ export class FeedReader {
                 const newGuids = [];
                 log.debug(`Found ${feed.items.length} entries in ${url}`);
                 for (const item of feed.items) {
-                    const guid = item.guid || item.id || item.link || item.title;
+                    // Find the first guid-like that looks like a string.
```
jaller94, line 249: Being a string doesn't make a value "guid-like".

jaller94 (approved):
Fair improvement, if the suggestion is accepted. :+1:

src/feeds/FeedReader.ts
Half-Shot: I would hash these before putting in memory. Someone could be a jerk and fill it with MBs of string, and then we'd put MBs into memory :)

Half-Shot (approved):

matthewza:
Good Job !

PR #691, opened 2023-03-27
Ensure joined to room when checking widget API permissions.

Half-Shot:
Soft dependency on https://github.com/matrix-org/matrix-appservice-bridge/pull/465

This should fix the cases where users see an error due to the invite not quite succeeding. This works by:

- Always ensuring we're joined to a room before testing permissions
- Making up to 10 (spaced by a second) checks for connections before giving up.
- Ensuring we log errors as JSON so we actually can check why we're failing.

### DIFF
--- a/src/provisioning/api.ts
+++ b/src/provisioning/api.ts
@@ -1,4 +1,4 @@
-import { Intent, MembershipEventContent, PowerLevelsEventContent } from "matrix-bot-sdk";
+import { Intent, MatrixError, MembershipEventContent, PowerLevelsEventContent } from "matrix-bot-sdk";
 import { ApiError, ErrCode } from "../api";
 import { Logger } from "matrix-appservice-bridge";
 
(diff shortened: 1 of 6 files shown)


src/provisioning/api.ts
```diff
@@ -33,6 +35,9 @@ export async function assertUserPermissionsInRoom(userId: string, roomId: string
             throw new ApiError("Bot is not joined to the room.", ErrCode.NotInRoom);
         }
     } catch (ex) {
+        if (ex instanceof MatrixError && ex.errcode === "M_FORBIDDEN") {
+            throw new ApiError(`User ${intent.userId} is not invited to the room.`, ErrCode.NotInRoom);
+        }
         if (isNotFoundError(ex)) {
```
justinbot, line 41: Should this error be checked in the same way as above (or vice-versa)?

src/provisioning/api.ts
```diff
@@ -25,6 +25,8 @@ export interface GetConnectionsResponseItem<Config = object, Secrets = object> e
 const log = new Logger("Provisioner.api");
 
 export async function assertUserPermissionsInRoom(userId: string, roomId: string, requiredPermission: "read"|"write", intent: Intent) {
+    // Always do an ensureJoined to clear any possible invites.
+    await intent.ensureJoined(roomId);
```
justinbot, line 29: This seems like it's going to happen too late -- we check if there is a bot user in the room before trying to assert permissions.

https://github.com/matrix-org/matrix-hookshot/blob/8071243696a39e7fa62d1fc7658bdfec46a3269a/src/Widgets/BridgeWidgetApi.ts#L152-L153

https://github.com/matrix-org/matrix-hookshot/blob/8071243696a39e7fa62d1fc7658bdfec46a3269a/src/Widgets/BridgeWidgetApi.ts#L98-L104

justinbot (approved):
Ensuring a bot is joined looks okay as a recovery step -- of course it would be ideal not to miss joins in the first place 🙂

PR #695, opened 2023-03-29
Allow users to import other people's go-neb services

tadzik:
This requires us to guess what these other people's MXIDs were, so we scroll through the list of room members and make educated guesses about which of them are Scalar+go-neb bots, and which users they were set up by.

Addresses https://github.com/vector-im/element-integration-manager/issues/57.

I wasn't able to test it on any live services yet, only on homebrew Scalar-like bots and service IDs. Needs testing in the wild.

### DIFF
--- a/changelog.d/696.feature
+++ b/changelog.d/696.feature
@@ -1,0 +1,1 @@
+Add support for push events on Github repo connections.
--- a/docs/usage/room_configuration/github_repo.md
+++ b/docs/usage/room_configuration/github_repo.md
@@ -65,6 +65,7 @@ the events marked as default below will be enabled. Otherwise, this is ignored.
   - pull_request.opened *
   - pull_request.ready_for_review *
   - pull_request.reviewed *
+- push
 - release *
   - release.created *
   - release.drafted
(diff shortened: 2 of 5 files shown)


Half-Shot (approved):
Seems pretty reasonable

src/Widgets/BridgeWidgetApi.ts
```diff
@@ -90,7 +91,16 @@ export class BridgeWidgetApi extends ProvisioningApi {
 
         const botUser = await this.getBotUserInRoom(roomId);
         await assertUserPermissionsInRoom(req.userId, roomId, "read", botUser.intent);
-        const connections = await this.goNebMigrator.getConnectionsForRoom(roomId, req.userId);
+
+        const userIds = [req.userId];
+
+        if ((await getUserPLInRoom(req.userId, roomId, botUser.intent)) == 100) {
```
justinbot, line 97: Seems slightly arbitrary to add another PL restriction on top of whatever is necessary to create a connection?

Half-Shot: I feel like anyone with the power to create connections should be able to read them from Go-NEB, that matches the hookshot logic. It's permissive, but not dangerously so.

Half-Shot (approved):
Still good!

PR #696, opened 2023-03-29
Support GitHub push events in repo connections

Half-Shot:
I've even gone and revitalized the concept of storing extra keys in the event body, so you can inspect which commits etc.

### DIFF
--- a/src/Connections/GithubRepo.ts
+++ b/src/Connections/GithubRepo.ts
@@ -25,7 +25,7 @@ import { GitHubIssueConnection } from "./GithubIssue";
 import { BridgeConfigGitHub } from "../Config/Config";
 import { ApiError, ErrCode, ValidatorApiError } from "../api";
 import { PermissionCheckFn } from ".";
-import { MinimalGitHubIssue, MinimalGitHubRepo } from "../libRs";
+import { GitHubRepoMessageBody, MinimalGitHubIssue, MinimalGitHubRepo } from "../libRs";
 import Ajv, { JSONSchemaType } from "ajv";
 import { HookFilter } from "../HookFilter";
 import { GitHubGrantChecker } from "../Github/GrantChecker";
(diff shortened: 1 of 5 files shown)


src/Connections/GithubRepo.ts
```diff
@@ -1259,6 +1277,29 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
...
+                commits: event.commits.map(c => c.id),
+                pusher: `${event.pusher.name} <${event.pusher.email}>`,
+                ref: event.ref,
+                base_ref: event.base_ref,
+            },
+            msgtype: "m.notice",
+            body: content,
+            formatted_body: md.render(content),
+            format: "org.matrix.custom.html",
+        } as IPushEventContent);
```
justinbot, line 1300: FWIW using `as` is an assertion which skips type checking.

It would get type checking for the interface if this was defined like:
```ts
const eventContent: IPushEventContent = {
    ...
```

src/Connections/GithubRepo.ts
```diff
@@ -1259,6 +1277,29 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
         }
     }
 
+    public async onPush(event: PushEvent) {
+        if (this.hookFilter.shouldSkip('push')) {
+            return;
+        }
+    
+        const content = `**${event.sender.login}** pushed [${event.commits.length} commit${event.commits.length === 1 ? '' : 's'}](${event.compare}) in ${event.ref} for ${event.repository.full_name}`;
```
justinbot, line 1285: Maybe a slight change of formatting?

**justinbot** push [2 commits](#) in justinbot/test for matrix-org/matrix-hookshot

**justinbot** push [2 commits](#) to `justinbot/test` on matrix-org/matrix-hookshot

### 
```suggestion
        const content = `**${event.sender.login}** pushed [${event.commits.length} commit${event.commits.length === 1 ? '' : 's'}](${event.compare}) to `${event.ref}` on ${event.repository.full_name}`;
```

justinbot (approved):
A few comments but looks good

src/Connections/GithubRepo.ts
Half-Shot: Oh...I've been doing this horribly wrong for ages.

PR #697, opened 2023-03-29
Add support for showing GitHub issue comments

Half-Shot:
Another Go-NEB parity feature with https://github.com/vector-im/element-integration-manager/issues/60

image

### DIFF
--- a/src/Bridge.ts
+++ b/src/Bridge.ts
@@ -231,11 +231,11 @@ export class Bridge {
             this.github?.onInstallationRemoved(data.data);
         });
 
-        this.bindHandlerToQueue<GitHubWebhookTypes.IssueCommentCreatedEvent, GitHubIssueConnection>(
+        this.bindHandlerToQueue<GitHubWebhookTypes.IssueCommentCreatedEvent, GitHubIssueConnection|GitHubRepoConnection>(
             "github.issue_comment.created",
             (data) => {
                 const { repository, issue, owner } = validateRepoIssue(data);
-                return connManager.getConnectionsForGithubIssue(owner, repository.name, issue.number).filter(c => c instanceof GitHubIssueConnection) as GitHubIssueConnection[];
+                return connManager.getConnectionsForGithubIssue(owner, repository.name, issue.number);
             },
             (c, data) => c.onIssueCommentCreated(data),
         );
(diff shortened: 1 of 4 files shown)


src/Connections/GithubRepo.ts
```diff
@@ -867,6 +872,24 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
         });
     }
 
+    public async onIssueCommentCreated(event: IssueCommentCreatedEvent) {
+        if (this.hookFilter.shouldSkip('issue.comment.created', 'issue.comment', 'issue') || !this.matchesLabelFilter(event.issue)) {
+            return;
+        }
+    
+        let message = `**${event.comment.user.login}** [commented](${event.issue.html_url}) on [${event.repository.full_name}#${event.issue.number}](${event.issue.html_url})  `;
+        message += "\n " + emoji.emojify(event.comment.body);
```
justinbot, line 881: Can this be inline quoted? I think it would make messages easier to read.

src/Bridge.ts
```diff
@@ -231,11 +231,11 @@ export class Bridge {
             this.github?.onInstallationRemoved(data.data);
         });
 
-        this.bindHandlerToQueue<GitHubWebhookTypes.IssueCommentCreatedEvent, GitHubIssueConnection>(
+        this.bindHandlerToQueue<GitHubWebhookTypes.IssueCommentCreatedEvent, GitHubIssueConnection|GitHubRepoConnection>(
             "github.issue_comment.created",
             (data) => {
                 const { repository, issue, owner } = validateRepoIssue(data);
-                return connManager.getConnectionsForGithubIssue(owner, repository.name, issue.number).filter(c => c instanceof GitHubIssueConnection) as GitHubIssueConnection[];
+                return connManager.getConnectionsForGithubIssue(owner, repository.name, issue.number);
```
justinbot, line 238: I'm slightly confused -- what is the difference between this feature and the issue comment handler which already exists?

https://github.com/matrix-org/matrix-hookshot/blob/2b5d4640c8ae288002e630e718425cdc409ecb18/src/Connections/GithubIssue.ts#L188-L233

justinbot (approved):
Looks good, just a couple minor things.

src/Bridge.ts
Half-Shot: The other one works in GitHub issue rooms as oppose to repo rooms, and it's quite a bit different in that it uses actual puppet users rather than a single bot user.

src/Connections/GithubRepo.ts
Half-Shot: Yeah, we can try that.

PR #701, opened 2023-04-03
Add `hookshot_` prefix on hookshot specific metrics

### LINKED ISSUE #699: All hookshot-specific metrics should start with hookshot_
We've got some like the feeds https://matrix-org.github.io/matrix-hookshot/3.1.1/metrics.html#feed which aren't prefix'd.

justinbot:
Ensures all metrics have a `hookshot_` prefix and updates the included dashboards.
This includes the `prom-client` [default metrics](https://github.com/siimon/prom-client#default-metrics).

Fixes #699


src/Metrics.ts
```diff
@@ -18,21 +18,22 @@ export class Metrics {
     public readonly notificationsServiceUp = new Gauge({ name: "hookshot_notifications_service_up", help: "Is the notification service up or down", labelNames: ["service"], registers: [this.registry]});
     public readonly notificationsWatchers = new Gauge({ name: "hookshot_notifications_watchers", help: "Number of notifications watchers running", labelNames: ["service"], registers: [this.registry]});
 
-    private readonly matrixApiCalls = new Counter({ name: "matrix_api_calls", help: "The number of Matrix client API calls made", labelNames: ["method"], registers: [this.registry]});
-    private readonly matrixApiCallsFailed = new Counter({ name: "matrix_api_calls_failed", help: "The number of Matrix client API calls which failed", labelNames: ["method"], registers: [this.registry]});
+    private readonly matrixApiCalls = new Counter({ name: "hookshot_matrix_api_calls", help: "The number of Matrix client API calls made", labelNames: ["method"], registers: [this.registry]});
```
Half-Shot, line 21: I wasn't sure about this one since we use it in other bridges, and it might break compat with other dashes (similar to the node exporter default metrics)

Half-Shot (changes requested):
Looks good, two thoughts

src/Metrics.ts
justinbot: FWIW this added a prefix on the default metrics as well using the [`prefix` parameter](https://github.com/siimon/prom-client#default-metrics).

I thought the goal was to make sure all the metrics are namespaced.
If we don't actually want to namespace metrics this way/ concerned about compatibility maybe we shouldn't bother?

Half-Shot: Yep, that's the idea. Anything common between bridges should be the same so people can reuse dashes (there is a "job" parameter provided by prometheus that allows you to distinguish if you need to)

Half-Shot:
> I don't really know what this metrics docs check is doing :/

Ah, we document all the metrics we provide and it requires you to manually run the script to update the docs when we change the metrics.

Half-Shot (approved):

Issue #706, opened 2023-04-11
Error fetching feed URL: input.replace is not a function


Half-Shot:
Looks dodgy, we should investigate.

PR #708, opened 2023-04-11
Don't process feed/item titles that aren't actually strings

tadzik:
Another case of empty-tag with attributes, these would occasionally parse into non-string objects and make us crash when trying to stripHtml on them.

Fixes https://github.com/matrix-org/matrix-hookshot/issues/706


Half-Shot (approved):
Looks good, worthy of a test?

Half-Shot (approved):
Looks quite nice!

PR #709, opened 2023-04-11
Use `rss` crate for RSS feed parsing

Half-Shot:
This obviously requires us to do a lot of testing prior to merging, but I'm curious to see how it performs. We've noticed a ton of incompatibility with the current RSS library, and so it would be nice to try a different one. The [rusty one](https://crates.io/crates/rss) looks really well supported and popular, and it was mostly a drop in for us.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "rust-analyzer.linkedProjects": [
+        "./Cargo.toml"
+    ]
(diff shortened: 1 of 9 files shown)


AndrewFerr (approved):
LGTM -- I can't speak too much about the Rust code, but at a glance it looks correct; and the overall design of this looks fine :+1:

PR #714, opened 2023-04-14
Allow usage of the SQLite-based crypto store

AndrewFerr:
Fixes https://github.com/matrix-org/matrix-hookshot/issues/609

### DIFF
--- a/docs/advanced/encryption.md
+++ b/docs/advanced/encryption.md
@@ -12,9 +12,10 @@ Hookshot supports end-to-bridge encryption via [MSC3202](https://github.com/matr
 
 In order for hookshot to use encryption, it must be configured as follows:
 - The `experimentalEncryption.storagePath` setting must point to a directory that hookshot has permissions to write files into. If running with Docker, this path should be within a volume (for persistency).
+- If hookshot had ever been started with a Sled-based crypto store, set `experimentalEncryption.useLegacySledStore` to `true`. Otherwise, omit that setting (or set it to `false`) to use the SQLite-based crypto store.
 - [Redis](./workers.md) must be enabled. Note that worker mode is not yet supported with encryption, so `queue.monolithic` must be set to `true`.
 
-If you ever reset your homeserver's state, ensure you also reset hookshot's encryption state. This includes clearing the `encryption.storagePath` directory and all worker state stored in your redis instance. Otherwise, hookshot may fail on start up with registration errors.
+If you ever reset your homeserver's state, ensure you also reset hookshot's encryption state. This includes clearing the `experimentalEncryption.storagePath` directory and all worker state stored in your redis instance. Otherwise, hookshot may fail on start up with registration errors.
 
 Also ensure that hookshot's appservice registration file contains every line from `registration.sample.yml` that appears after the `If enabling encryption` comment. Note that changing the registration file may require restarting the homeserver that hookshot is connected to.
 
--- a/docs/advanced/encryption.md
+++ b/docs/advanced/encryption.md
@@ -11,8 +11,9 @@ Hookshot supports end-to-bridge encryption via [MSC3202](https://github.com/matr
 ## Enabling encryption in Hookshot
 
 In order for hookshot to use encryption, it must be configured as follows:
-- The `experimentalEncryption.storagePath` setting must point to a directory that hookshot has permissions to write files into. If running with Docker, this path should be within a volume (for persistency).
-- If hookshot had ever been started with a Sled-based crypto store, set `experimentalEncryption.useLegacySledStore` to `true`. Otherwise, omit that setting (or set it to `false`) to use the SQLite-based crypto store.
+- The `experimentalEncryption.storagePath` setting must point to a directory that hookshot has permissions to write files into. If running with Docker, this path should be within a volume (for persistency). Hookshot uses this directory for its crypto store (i.e. long-lived state relating to its encryption keys).
+    - By default, the crypto store uses an SQLite-based format. To use the legacy Sled-based format, set `experimentalEncryption.useLegacySledStore` to `true`, though this is not expected to be necessary and will be removed in a future release.
+    - Once a crypto store of a particular type (SQLite or Sled) has been initialized, its files must not be modified, and hookshot cannot be configured to use another crypto store of the same type as one it has used before. If a crypto store's files get lost or corrupted, hookshot may fail to start up, or may be unable to decrypt command messages. To fix such issues, stop hookshot, then reset its crypto store by running `yarn start:resetcrypto`.
 - [Redis](./workers.md) must be enabled. Note that worker mode is not yet supported with encryption, so `queue.monolithic` must be set to `true`.
 
 If you ever reset your homeserver's state, ensure you also reset hookshot's encryption state. This includes clearing the `experimentalEncryption.storagePath` directory and all worker state stored in your redis instance. Otherwise, hookshot may fail on start up with registration errors.
--- a/src/App/ResetCryptoStore.ts
+++ b/src/App/ResetCryptoStore.ts
@@ -6,7 +6,7 @@ import { LogService, MatrixClient } from "matrix-bot-sdk";
 import { getAppservice } from "../appservice";
 import BotUsersManager from "../Managers/BotUsersManager";
 
-const log = new Logger("App");
+const log = new Logger("ResetCryptoStore");
 
 async function start() {
     const configFile = process.argv[2] || "./config.yml";
@@ -63,12 +63,12 @@ async function start() {
             try {
                 await userStorage.storeValue("accessToken", "");
                 log.info(`Deleted access token for ${botUser.userId}`);
-            } catch (err: unknown) {
-                log.error(`Failed to delete access token for ${botUser.userId}: ${err}`);
+            } catch (ex: unknown) {
+                log.error(`Failed to delete access token for ${botUser.userId}`, ex);
             }
 
-        } catch (err: unknown) {
-            log.error(`Failed to log out crypto device for ${botUser.userId}: ${err}`);
+        } catch (ex: unknown) {
+            log.error(`Failed to log out crypto device for ${botUser.userId}`, ex);
         }
     }
 
@@ -76,8 +76,8 @@ async function start() {
         try {
             await rm(config.encryption.storagePath, { recursive: true, force: true });
             log.info("Removed crypto store from disk");
-        } catch (err) {
-            log.error("Failed to remove crypto store from disk");
+        } catch (ex) {
+            log.error("Failed to remove crypto store from disk", ex);
         }
     }
     // Process hangs without this, maybe because of a leftover promise
--- a/src/App/ResetCryptoStore.ts
+++ b/src/App/ResetCryptoStore.ts
@@ -1,0 +1,89 @@
+import { rm } from "fs/promises"
+
+import { BridgeConfig, parseRegistrationFile } from "../Config/Config";
+import { Logger } from "matrix-appservice-bridge";
+import { LogService, MatrixClient } from "matrix-bot-sdk";
+import { getAppservice } from "../appservice";
+import BotUsersManager from "../Managers/BotUsersManager";
+
+const log = new Logger("App");
+
+async function start() {
+    const configFile = process.argv[2] || "./config.yml";
+    const registrationFile = process.argv[3] || "./registration.yml";
+    const config = await BridgeConfig.parseConfig(configFile, process.env);
+    const registration = await parseRegistrationFile(registrationFile);
+    Logger.configure({
+        console: config.logging.level,
+        colorize: config.logging.colorize,
+        json: config.logging.json,
+        timestampFormat: config.logging.timestampFormat
+    });
+    LogService.setLogger(Logger.botSdkLogger);
+
+    const {appservice, storage, cryptoStorage} = getAppservice(config, registration);
+    if (!cryptoStorage) {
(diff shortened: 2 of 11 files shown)


docs/advanced/encryption.md
```diff
@@ -12,9 +12,10 @@ Hookshot supports end-to-bridge encryption via [MSC3202](https://github.com/matr
 
 In order for hookshot to use encryption, it must be configured as follows:
 - The `experimentalEncryption.storagePath` setting must point to a directory that hookshot has permissions to write files into. If running with Docker, this path should be within a volume (for persistency).
+- If hookshot had ever been started with a Sled-based crypto store, set `experimentalEncryption.useLegacySledStore` to `true`. Otherwise, omit that setting (or set it to `false`) to use the SQLite-based crypto store.
```
Half-Shot, line 15: I am wondering if this is breaking for the present day sled users, or if we care given experimental flags. We should also explain we don't have a migration path for Sled-ers.

Half-Shot, line 15: Part of me wonders if this should be a `storePath` with a `sqlite://file` format, like we do for our other bridges. We may end up doing psql formats too, and then you'll have `useLegacySledStore`, `useLegacySqliteStore` and `postgresUri` :D

src/appservice.ts
```diff
@@ -17,7 +17,11 @@ export function getAppservice(config: BridgeConfig, registration: IAppserviceReg
         storage = new MemoryStorageProvider();
     }
 
-    const cryptoStorage = config.encryption?.storagePath ? new RustSdkAppserviceCryptoStorageProvider(config.encryption.storagePath) : undefined;
+    const cryptoStorage = config.encryption?.storagePath ?
+        new RustSdkAppserviceCryptoStorageProvider(
```
Half-Shot, line 21: should we loggy that we're setting up crypto (with the path)?

Half-Shot (changes requested):
(I'm a dumb and didn't review comment the first two comments, but anyway, some :canned_food: for :thought_balloon: )

docs/advanced/encryption.md
AndrewFerr: What makes this less straightforward is that the `storagePath` is not a path to a single file, but to a path to a _directory_ in which the bot-sdk saves crypto-related files for every appservice user.

At the time of writing, this includes the following files for each appservice user:
- a JSON file created by the bot-sdk for some basic info about the user (currently just the ID of their crypto-enabled device, which rooms they're in, and some crypto-relevant metadata for those rooms) 
- the SQLite (or Sled :slightly_smiling_face:) crypto store file created by the rust-sdk, for the meat of the user's crypto state

Notably, the crypto store files created by the rust-sdk are per-user, not per-appservice.

These files are organized in subdirectories within `storagePath`, one for each appservice user (as well as for a "default" user representing the appservice itself).

Given this, it seems like the bot-sdk (or downstream apps like Hookshot) can't use a single file / connection URL for the whole crypto store until both the bot-sdk & rust-sdk support per-appservice stores.

Even if the rust-sdk were to support Postgresql and manage all appservice users in the same database, the bot-sdk would still need some refactoring to not need those per-user JSON files it creates.

src/appservice.ts
AndrewFerr: Sure, couldn't hurt

docs/advanced/encryption.md
AndrewFerr: In short: I'd rather incur the potential inconvenience of (yet another) migration path for when & if Postgresql is supported, instead of potentially making a wrong early guess as to how the rust-sdk & bot-sdk might work with a single file / database for crypto state.

AndrewFerr: Better yet: Hookshot should check if a crypto file of an unselected format already exists, and error-out if so.

Half-Shot: Ah, right I forgot we have *both* a JSON store and a SQLite store for each user. That's more of an issue. Okay, suggest when postgres comes along we treat it as a part of a larger refactor to reduce file usage.

AndrewFerr: On second thought, the rust-sdk doesn't provide a way to check if a crypto store of a particular type already exists (and trying to check manually isn't straightforward, since the rust-sdk doesn't expose the filenames of its crypto stores), so it's not really worth the effort to be so diligent about it & we can just get away with clearly documenting the importance of this config setting.

AndrewFerr: Looks like I overcomplicated this all along: migrating to a new store type is painless; only migrating back to a previously-used store type (after having deleted/moved the original crypto store files) is troublesome.

AIUI the trouble in question is Hookshot crashing on startup if it tries to use a different E2EE device key than it used previously for that device (i.e. login session).  Switching to a never-before-used crypto store type doesn't hit this issue, since changing the store type also gives Hookshot bots a new device entirely, along with a new device key.

As unlikely as this kind of back-and-forth migration should be in practice, I'll document it just in case.

AndrewFerr:
Barring review, this PR should now be done, other than updating the bot-sdk dependency (once it becomes available).

The yarn script for resetting crypto state should help fix "stuck" Hookshots that crash on startup with E2EE errors, or fail to decrypt messages even after resetting room keys (like with `/discardsession` or a page refresh in Element Web).  The scenarios that cause those issues shouldn't happen during normal usage, but they're possible; having a command to get things back in order should come in handy.

src/App/ResetCryptoStore.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    if (config.encryption?.storagePath) {
+        try {
+            await rm(config.encryption.storagePath, { recursive: true, force: true });
+            log.info("Removed crypto store from disk")
+        } catch (err) {
+            log.error("Failed to remove crypto store from disk")
+        }
+    }
+    // Process hangs without this, maybe because of a leftover promise
```
Half-Shot, line 83: *Suspicious noises*

src/App/ResetCryptoStore.ts
```diff
@@ -0,0 +1,89 @@
...
+
+            try {
+                await Promise.resolve(userStorage.storeValue("accessToken", ""))
+                log.info(`Deleted access token for ${botUser.userId}`)
+            } catch (err: unknown) {
+                log.error(`Failed to delete access token for ${botUser.userId}: ${err}`)
+            }
+
+        } catch (err: unknown) {
+            log.error(`Failed to log out crypto device for ${botUser.userId}: ${err}`)
```
Half-Shot, line 71: We *tend* to leave the error as an argument parameter so it formats better. If memory serves by default this just logs the message without the stack?

src/App/ResetCryptoStore.ts
```diff
@@ -0,0 +1,89 @@
...
+
+            const client = new MatrixClient(config.bridge.url, accessToken, userStorage, userCryptoStorage);
+            await client.doRequest("POST", "/_matrix/client/v3/logout", {
+                user_id: botUser.userId,
+                "org.matrix.msc3202.device_id": deviceId,
+            });
+            log.info(`Logged out crypto device for ${botUser.userId}`)
+
+            try {
+                await Promise.resolve(userStorage.storeValue("accessToken", ""))
```
Half-Shot, line 64: errrr?

src/App/ResetCryptoStore.ts
```diff
@@ -0,0 +1,89 @@
...
+    const botUsersManager = new BotUsersManager(config, appservice);
+
+    for (const botUser of botUsersManager.botUsers) {
+        try {
+            const userStorage = storage.storageForUser?.(botUser.userId)
+            if (!userStorage) {
+                log.warn(`No storage for ${botUser.userId}`);
+                continue;
+            }
+            const accessToken = await Promise.resolve(userStorage?.readValue("accessToken"));
```
Half-Shot, line 39: Hm? Oh, is this a Promise<T>|T type? You should be able to await it without the Promise.resolve.

src/App/ResetCryptoStore.ts
```diff
@@ -0,0 +1,89 @@
+import { rm } from "fs/promises"
+
+import { BridgeConfig, parseRegistrationFile } from "../Config/Config";
+import { Logger } from "matrix-appservice-bridge";
+import { LogService, MatrixClient } from "matrix-bot-sdk";
+import { getAppservice } from "../appservice";
+import BotUsersManager from "../Managers/BotUsersManager";
+
+const log = new Logger("App");
```
Half-Shot, line 9:
```suggestion
const log = new Logger("ResetCryptoStore");
```

Half-Shot (changes requested):
Rest of the changes look good, but a few things in the new script look suspect

src/App/ResetCryptoStore.ts
```diff
@@ -0,0 +1,89 @@
+import { rm } from "fs/promises"
```
Half-Shot, line 1: N.B. Unwritten rule here is `App` is for long running processes and we have a scripts/ directory for scripts.

src/App/ResetCryptoStore.ts
AndrewFerr: Your memory serves you right; I'll move this to a parameter.

src/App/ResetCryptoStore.ts
AndrewFerr: This paradigm is from the bot-sdk, for safely awaiting a function call that returns `SomeType|Promise<SomeType>` ([example](https://github.com/turt2live/matrix-bot-sdk/blob/main/src/appservice/Intent.ts#L148), and [another](https://github.com/turt2live/matrix-bot-sdk/blob/main/src/appservice/Intent.ts#L321)).

src/App/ResetCryptoStore.ts
AndrewFerr: Since this is a TS file, I'd like to run it with `ts-node` as is done for some of our other yarn scripts do, but most of them don't work.  For example, here's what I get from `yarn validate-config`:
```
yarn run v1.22.19
$ ts-node src/Config/Config.ts
<PATH_TO_HOOKSHOT>/matrix-hookshot/node_modules/ts-node/src/index.ts:944
          throw new TypeError(`${relative(cwd, fileName)}: Emit skipped`);
                ^
TypeError: ../matrix-bot-sdk/lib/index.js: Emit skipped
    at getOutput (<PATH_TO_HOOKSHOT>/matrix-hookshot/node_modules/ts-node/src/index.ts:944:17)
    at Object.compile (<PATH_TO_HOOKSHOT>/matrix-hookshot/node_modules/ts-node/src/index.ts:1243:30)
    at Module.m._compile (<PATH_TO_HOOKSHOT>/matrix-hookshot/node_modules/ts-node/src/index.ts:1370:30)
    at Module._extensions..js (node:internal/modules/cjs/loader:1209:10)
    at Object.require.extensions.<computed> [as .js] (<PATH_TO_HOOKSHOT>/matrix-hookshot/node_modules/ts-node/src/index.ts:1374:12)
    at Module.load (node:internal/modules/cjs/loader:1033:32)
    at Function.Module._load (node:internal/modules/cjs/loader:868:12)
    at Module.require (node:internal/modules/cjs/loader:1057:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (<PATH_TO_HOOKSHOT>/matrix-hookshot/src/Config/Config.ts:3:1)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Is there some prerequisite step needed for `ts-node` scripts to work?

Half-Shot: Errrm, no. I've never seen that before :worried: . Is this related to our fork?

src/App/ResetCryptoStore.ts
AndrewFerr: [From MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#description):
> ```javascript
> await expression
> ```
> ...
> The expression is resolved in the same way as [Promise.resolve()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve): it's always converted to a native Promise and then awaited.

So the `Promise.resolve` here is indeed redundant.

src/App/ResetCryptoStore.ts
AndrewFerr: Your suggestion to use [why-is-node-still-running](https://www.npmjs.com/package/why-is-node-still-running) did reveal the cause of the hang after all!: the Redis store.  Closing it lets the process terminate cleanly.

src/App/ResetCryptoStore.ts
AndrewFerr: No; it turns out that Hookshot's `ts-node` setup is improper.  It's broken due to `scripts/.*ts` files not being in the `tsconfig.json`'s `rootDir`.

See: https://github.com/TypeStrong/ts-node/discussions/1595

I gave a shot at fixing it, but it's not quite taking.  This same issue affects `scripts/definition-fixer.ts` too, so IMO this is worth fixing in a separate changelist, as long as it's fine for the crypto reset script to stay where it is for now (note that even after being moved, it would be invoked by the same `yarn` command as it is now).

Half-Shot: Yeah, given we alias a yarn command I'm not super hung up on it. Though we should fix the improper setup :)

Half-Shot (approved):
I'm happy once CI passes

src/App/ResetCryptoStore.ts
AndrewFerr: https://github.com/matrix-org/matrix-hookshot/issues/722

Issue #541, opened 2022-10-18
Give ability to disable error reporting into matrix rooms


etkecc:
Please, give ability to disable errors like

```
Error fetching https://github.com/Half-Shot/matrix-appservice-discord/tags.atom: timeout of 10000ms exceeded
```

in the matrix rooms

HarHarLinks:
Related #493

I've posted some related ideas to improve error behaviour instead of disabling it outright to chat, quoted here for convenience:

> alternatively to showing the rss error on first timeout(?) and increasing the timeout from 10s to 30s, it could only show it on consecutive poll timeouts?
so if it times out once then it fails silently, but if $interval minutes later it times out again, then it sends an error to matrix.
the argument being that feeds should not be so time sensitive since you probably don't expect them to be absolutely realtime

> Maybe the bridge could consolidate multiple RSS timeout errors into one message instead of repeating it for potentially every feed a room is subscribed to?

HarHarLinks:
It would be neat to have more error resilience on one hand, but also error reporting only to admin rooms for example on the other.

Half-Shot:
Yeah, suspect you are both right. We should reduce the amount of noise from hookshot in the rooms, but also try to fix the timeouts.

vranki:
I'd like to have timeout error disabled by default. In current form it just spams room if there is a random glitch. If the feed URL can't be reached, it could be displayed in the widget.

HarHarLinks:
> If the feed URL can't be reached, it could be displayed in the widget.

i get that y'all don't want to get spammed. personally I bumped the timeout from 10s to 30s and it's fine apart from very few exceptions.

if you disable errors in chat, how will you notice that a feed broke? do you check the widget every now and then manually? or do you just not care and accept that feeds may silently disappear/break?

SISheogorath:
I would suggest to change the threshold for the errors. A single timeout doesn't mean anything. If a feed couldn't be fetched for a week, this is a legitimate concern. But being unsuccessful once? Nope, please don't, a matrix room is not a (de-)centralised log system. This only leads to alert fatigue.

Half-Shot:
Hiya, sorry about the noise. I had thought this *did* only alert on several failures although I think that logic may not be as smooth as we hoped. I'll put in something to ensure it doesn't make so much noise. Please hang tight.

HarHarLinks:
This appears to have been resolved with a very black and white decision users now have to make Is there another more nuanced option planned?

PR #716, opened 2023-04-18
Add option to enable/disable feed failure notices.

Half-Shot:
Fixes https://github.com/matrix-org/matrix-hookshot/issues/541
Fixes https://github.com/vector-im/element-integration-manager/issues/63

This disables the notices by default when a feed fails, and adds a checkbox to opt in.

This also fixes a problem with the room configs not visibly changing when state is saved, by disabling the save/delete buttons while a save is in progress.

### DIFF
--- a/src/Notifications/GitLabWatcher.ts
+++ b/src/Notifications/GitLabWatcher.ts
@@ -29,8 +29,5 @@ export class GitLabWatcher extends EventEmitter implements NotificationWatcherTa
 
     private async getNotifications() {
         log.info(`Fetching events from GitLab for ${this.userId}`);
-        const events = await this.client.getEvents({
-            after: new Date(this.since)
-        });
     }
 }
--- a/web/components/roomConfig/FeedsConfig.tsx
+++ b/web/components/roomConfig/FeedsConfig.tsx
@@ -28,7 +28,7 @@ const FeedRecentResults: FunctionComponent<{item: FeedResponseItem}> = ({ item }
 }
 const DOCUMENTATION_LINK = "https://matrix-org.github.io/matrix-hookshot/latest/setup/feeds.html#feed-templates";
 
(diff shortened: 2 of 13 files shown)


web/components/roomConfig/FeedsConfig.tsx
```diff
@@ -58,14 +62,16 @@ const ConnectionConfiguration: FunctionComponent<ConnectionConfigurationProps<Se
         <InputField visible={true} label="Label" noPadding={true}>
             <input ref={labelRef} disabled={!canSave} type="text" value={existingConnection?.config.label} />
         </InputField>
-        <InputField visible={true} label="Template" noPadding={true}>
+        <InputField visible={onlyVisibleOnExistingConnection} label="Send a notice on read failure" noPadding={true}>
```
AndrewFerr, line 65: It would be nice if these extra controls (the read failure & template sections) also appeared for the feed to be added (instead of only for already-added feeds) so that they can be set without having to add the feed first.  But that can be done in a later PR (if we even want that).

src/Notifications/GitLabWatcher.ts
```diff
@@ -29,8 +29,5 @@ export class GitLabWatcher extends EventEmitter implements NotificationWatcherTa
 
     private async getNotifications() {
         log.info(`Fetching events from GitLab for ${this.userId}`);
-        const events = await this.client.getEvents({
```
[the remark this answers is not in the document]
Half-Shot: This code has done nothing for ages and we've never really gone back. This was just fixing a linting issue.

web/components/roomConfig/FeedsConfig.tsx
Half-Shot: It's basically intentional to avoid having the UI overwhelm the user. The create new feed UI took up so much verical space with this visible that I felt it would be better to show just the URL and Label and allow users to adjust to taste afterwards.

AndrewFerr (approved):

HarHarLinks:
I hope this will be announced as a (UX-)breaking change?

PR #718, opened 2023-04-18
Fix case where GitHub repos aren't connectable if the GitHub App was manually approved

### LINKED ISSUE #711: Hookshot is not informed when a GitHub App installation goes from pending -> active
GitHub do not send webhooks informing you of the install in these cases (at least, not in the instance I checked), so we probably need to manually refresh the cache.

You can fix this by restarting, but that's not very practical on big installs.

Half-Shot:
Believe this fixes #711 

Instead of relying on our internal cache for GitHub apps, let's just check on provisionConnection to see if GitHub has access to a repo.

### DIFF
--- a/changelog.d/718.bugfix
+++ b/changelog.d/718.bugfix
@@ -1,0 +1,1 @@
+Fix cases of GitHub repos not being bridgable if the GitHub App had to be manually approved.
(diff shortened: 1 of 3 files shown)


AndrewFerr (approved):
The code looks fine, and I trust that it works as intended.

PR #721, opened 2023-04-21
Fallback to parsing feeds as atom format if rss format fails.

Half-Shot:
This should fix the new cases of feeds failing due to the new rust parser only handing RSS format. However, unsure if I like the pattern of: 

1. Parse XML
2. Look for RSS tag
3. If not found, parse XML again
4. Look for feed tag


AndrewFerr (approved):
Other than the trivial lint rule violation, this looks fine.  It would be nice to use some kind of polymorphism to avoid having to parse twice, but we can put this in as a known-working first step & consider refactoring it later.

PR #723, opened 2023-04-21
Fix hang in crypto reset script when E2EE is off

AndrewFerr:
Ensure that the bridge storage is always disconnected

---

Without this, `yarn start:resetcrypto` hangs if E2EE is disabled in the config file.


AndrewFerr:
IMO we can forego the changelog file here because this just fixes a tiny bug left behind in the not-yet-released [E2EE fix PR](https://github.com/matrix-org/matrix-hookshot/pull/714).

Half-Shot (approved):

PR #724, opened 2023-04-21
Support node 20, drop node 16

Half-Shot:
Inline with our changes on https://github.com/matrix-org/matrix-appservice-bridge/pull/466


.github/workflows/main.yml
```diff
@@ -16,18 +16,18 @@ jobs:
...
    runs-on: ubuntu-latest
    steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Use Node.js
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v3
       with:
-        node-version: 16
+        node-version-file: .node-version
```
justinbot, line 23: Looks like the `.node-version` file needs to be updated

Half-Shot: Should be up to date?

justinbot: Oh you're right, I just missed it 🙈

AndrewFerr (approved):
Should finally be ready!

tsconfig.json
```diff
@@ -25,6 +25,7 @@
     "emitDecoratorMetadata": true,
     // TODO: Type errors
     "useUnknownInCatchVariables": false,
+    "target": "es2021"
```
Half-Shot, line 28: This *shouldn't* be required if we're extending from tsconfig?

Half-Shot: I think we just need to update Typescript

AndrewFerr: > This shouldn't be required if we're extending from tsconfig?

True, but this is here to override the `"target"` inherited from `@tsconfig/node18/tsconfig.json`, which is `"es2022"`.  For some reason, using that value breaks a few of our CI checks: `config`, `metrics-docs`, and `test (18/20)`.  They fail when run locally, too.  Here's an example of a failed CI run:  https://github.com/matrix-org/matrix-hookshot/actions/runs/4811143763

Tweaking `package.json` to either revert any updates made in this PR, or update some other deps to their latest versions, doesn't help.

It also works to extend `@tsconfig/node16/tsconfig.json`, which sets `"target": "es2021"`, along with some other minor differences from `node18`.  It turns out the only breaking difference is `"target"`, so I figured it was a more future-facing change to to override `"target"` to its `node16` default and let the rest be inherited from `node18`.

> I think we just need to update Typescript

It was already updated to its latest version in 5632112b, which at the time of writing is still the latest version (^5.0.4).

Half-Shot: Solved in 4d87454

Half-Shot: Not solved, reverted, and #729 created.

PR #743, opened 2023-05-05
Show GitHub workflow errors in a thread below failure messages

### LINKED ISSUE #742: Emit the stage that failed and any error logs when a GitHub workflow fails
It's useful to be able to know at a glance why something failed. I don't think it would be too much work to figure it out :)

Half-Shot:
Fixes #742 

Nice and simple. Logs are quite a bit harder but folks should be able to click through to GitHub for that.

image

image

### DIFF
--- a/src/Connections/GithubRepo.ts
+++ b/src/Connections/GithubRepo.ts
@@ -325,6 +325,12 @@ const WORKFLOW_CONCLUSION_TO_NOTICE: Record<WorkflowRunCompletedEvent["workflow_
     stale: "completed, but is stale 🍞"
 }
 
+const WORKFLOW_JOB_TO_NOTICE: Record<string, string> = {
+    success: "✅",
+    failure: "failed ⚠️",
+    skipped: "skipped ⏭️"
+}
+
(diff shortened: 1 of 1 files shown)


src/Connections/GithubRepo.ts
```diff
@@ -1260,12 +1266,72 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
...
+        if (extraThreadContent) {
+            this.intent.sendEvent(this.roomId, {
+                "m.relates_to": {
+                    rel_type: "m.thread",
+                    event_id: matrixEvent,
+                    // Needed to prevent clients from showing these as actual replies
+                    is_falling_back: true,
+                    "m.in_reply_to": {
+                        event_id: matrixEvent,
+                    }
```
clokep, line 1326: Do you really want this to be a reply too or just part of the thread?

Half-Shot: I think we want the fallback to be a reply?

clokep: Well then, I had fully forgotten this was a thing. 👍

PR #746, opened 2023-05-08
fix: Make sure that the blockquote markdown for github issue comments are valid commonmark

MTRNord:
Fixes that blockquotes don't get rendered as such. Leading spaces in front of `>` are not valid in commonmark. So they get not properly converted, but instead they seem to get escaped.

This PR was made on the assumption that the lib used follows commonmark spec as closely as it claims, and this being the logical result for this. I didn't have a test env handy to actually validate this.

Signed-off-by: Marcel Radzio <support@nordgedanken.dev>

### DIFF
--- a/changelog.d/746.bugfix
+++ b/changelog.d/746.bugfix
@@ -1,0 +1,1 @@
+Fix Github comments not being rendered correctly as blockquotes.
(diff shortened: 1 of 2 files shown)


Half-Shot (approved):
Seems okay, usual changelogs and signoffs please :)

PR #754, opened 2023-05-17
Add Sentry support

Half-Shot:
We've been finding this useful in the IRC bridge arena, so let's try it in hookshot too.


config.sample.yml
```diff
@@ -157,6 +157,11 @@ widgets:
   publicUrl: https://example.com/widgetapi/v1/static/
   branding:
     widgetTitle: Hookshot Configuration
+sentry:
+  # (Optional) Add support for sentry.io tracing
```
justinbot, line 161: I don't think we're using tracing (https://docs.sentry.io/product/sentry-basics/tracing/), just error reporting?

src/Bridge.ts
```diff
@@ -772,17 +774,31 @@ export class Bridge {
...
 
+    private handleHookshotEvent<EventType, ConnType extends IConnection>(msg: MessageQueueMessageOut<EventType>, connection: ConnType, handler: (c: ConnType, data: EventType) => Promise<unknown>|unknown) {
+        Sentry.withScope((scope) => {
+            scope.setTransactionName('handleHookshotEvent');
+            scope.setTags({
+                eventType: msg.eventName,
+                roomId: connection.roomId,
+            });
+            scope.setContext("connection", {
+                id: connection.connectionId,
```
justinbot, line 785: Should this be a tag instead of / in addition to context? (Tags are searchable and context is not)

src/ListenerService.ts
```diff
@@ -74,6 +75,8 @@ export class ListenerService {
             listener.app.get("/live", (_, res) => res.send({ok: true}));
             listener.app.get("/ready", (_, res) => res.status(listener.resourcesBound ? 200 : 500).send({ready: listener.resourcesBound}));
 
+            // By default, Sentry only traces 500+ errors, which is what we want.
```
justinbot, line 78:
```suggestion
            // By default, Sentry only reports 500+ errors, which is what we want.
```

justinbot (approved):
A few comments but looks good

src/Bridge.ts
Half-Shot: The connectionId contains a unique ID, so I didn't want to saturate indexes.

justinbot: Ah that makes sense.

PR #757, opened 2023-05-18
Don't match issue comment notices on issue event type

Half-Shot:


src/Connections/GithubRepo.ts
```diff
@@ -907,7 +907,7 @@ export class GitHubRepoConnection extends CommandConnection<GitHubRepoConnection
     }
 
     public async onIssueCommentCreated(event: IssueCommentCreatedEvent) {
-        if (this.hookFilter.shouldSkip('issue.comment.created', 'issue.comment', 'issue') || !this.matchesLabelFilter(event.issue)) {
+        if (this.hookFilter.shouldSkip('issue.comment.created', 'issue.comment') || !this.matchesLabelFilter(event.issue)) {
```
Half-Shot, line 910: for ref, these are at different levels

image

PR #775, opened 2023-06-08
Add toggle for bridge queue config

Arkaniad:
With the [2.6.0](https://github.com/matrix-org/matrix-hookshot/releases/tag/2.6.0) release it looks like worker support with Redis was configured. With the Helm Chart using the default configuration, the values.yaml has a configured `queue` block since it's in the `config.sample.yml` file. 

Given that the chart doesn't yet deploy Redis, this results in a broken config by default unless `.Values.hookshot.config.queue` is set to `{}` as user-specified values will be merged on top of chart default values. 

This will also break user configurations if they set up shop using the `config.sample.yml` file as it exists prior to this commit.

We're just adding an `enabled` flag to the `queue` config key and defaulting it to false - That way Hookshot should never expect a Redis service to be available unless explicitly configured and enabled.

### DIFF
--- a/src/config/Config.ts
+++ b/src/config/Config.ts
@@ -37,7 +37,7 @@ export enum BridgePermissionLevel {
 interface BridgeConfigGitHubYAML {
     enterpriseUrl?: string;
     auth: {
-        id: number|string;
+        id: number | string;
         privateKeyFile: string;
     };
     webhook: {
@@ -58,7 +58,7 @@ interface BridgeConfigGitHubYAML {
 export class BridgeConfigGitHub {
     @configKey("Authentication for the GitHub App.", false)
     readonly auth: {
-        id: number|string;
+        id: number | string;
         privateKeyFile: string;
     };
     @configKey("Webhook settings for the GitHub app.", false)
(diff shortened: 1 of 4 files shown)


src/config/Config.ts
```diff
@@ -526,7 +529,7 @@ export class BridgeConfig {
     @hideKey()
     private readonly bridgePermissions: BridgePermissions;
 
-    constructor(configData: BridgeConfigRoot, env?: {[key: string]: string|undefined}) {
+    constructor(configData: BridgeConfigRoot, env?: { [key: string]: string | undefined }) {
         this.bridge = configData.bridge;
         assert.ok(this.bridge);
         this.github = configData.github && new BridgeConfigGitHub(configData.github);
```
Half-Shot, line 535: Would suggest modifying line 552 so that if `enabled` is false, we treat it as `{ monolithic: true}`

Half-Shot (changes requested):
Your system linter appears to have gone to town :p

src/config/Config.ts
Arkaniad: Handled in latest commit - I'm not a great JS dev, feel free to critique

PR #779, opened 2023-06-12
Enable feeds to be polled concurrently

### LINKED ISSUE #703: Run several feed fetches concurrently
We currently fetch feeds in a simple FIFO pattern but it's fairly dumb and so if you've got 10000 feeds, it's going to take a while. Also, a feed that just times out will clog up the whole process. We should instead run multiple feed checks concurrently.

A smarter solution would be to actually rank the responsiveness of feeds and the slow ones can be put into a slow queue, so as to not slow down the reasonably speedy ones.

Half-Shot:
Fixes #703 

Thereby speeding up the feed polling business by making it harder for one slow feed to block the queue. Future changes could make the polling concurrency more dynamic depending on load, or weight feeds into different queues so they roughly all chug along at the same pace.


bkil:
Will this fix #703 ?

src/feeds/FeedReader.ts
```diff
@@ -189,7 +190,9 @@ export class FeedReader {
         log.debug('Loaded feed URLs:', this.observedFeedUrls);
 
         void this.loadSeenEntries().then(() => {
-            return this.pollFeeds();
+            for (let workerIndex = 0; workerIndex < FEED_CONCURRENCY; workerIndex++) {
+                return this.pollFeeds(workerIndex);
```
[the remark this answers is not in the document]
Half-Shot: Should now work.

Half-Shot:
> Will this fix #703 ?

it will!

src/feeds/FeedReader.ts
```diff
@@ -189,7 +188,7 @@ export class FeedReader {
         log.debug('Loaded feed URLs:', this.observedFeedUrls);
 
         void this.loadSeenEntries().then(() => {
-            return this.pollFeeds();
+            Promise.all(Array.from({length: config.pollConcurrency}, (_, i) => this.pollFeeds(i)))
```
AndrewFerr, line 191: As long as nothing is awaiting/thenning on `Promise.all`, could this just be a loop of `this.pollFeeds` calls that neither returns nor awaits?
```suggestion
            for (let i = 0; i < config.pollConcurrency; i++) {
                this.pollFeeds(i);
            }
```

src/feeds/FeedReader.ts
```diff
@@ -365,40 +364,44 @@ export class FeedReader {
...
      */
-    public async pollFeeds(): Promise<void> {
-        log.debug(`Checking for updates in ${this.observedFeedUrls.size} RSS/Atom feeds`);
+    public async pollFeeds(workerId: number): Promise<void> {
+
+        // Update on each iteration
+        Metrics.feedsFailing.set({ reason: "http" }, this.feedsFailingHttp.size );
+        Metrics.feedsFailing.set({ reason: "parsing" }, this.feedsFailingParsing.size);
+        Metrics.feedsFailingDeprecated.set({ reason: "http" }, this.feedsFailingHttp.size );
+        Metrics.feedsFailingDeprecated.set({ reason: "parsing" }, this.feedsFailingParsing.size);
```
AndrewFerr, line 373: Since `this.pollFeed` is what updates these counts, does updating the metrics before the awaited call to `this.pollFeed` mean there's a risk of each worker making overlapping updates of the same values?

Should there be an async lock on making these updates?

src/feeds/FeedReader.ts
Half-Shot: Yeah, can do. The previous code returned a promise so I felt it was right to, but it's not like we're using it!

src/feeds/FeedReader.ts
Half-Shot: It should be fine. This section of code just checks the size of a Set and updates the metric each time any worker finishes, overlapping updates are fine because the workers are all updating the same shared value, so as long as it gets updated each time a worker finishes it will be correct.

AndrewFerr: It looks like they get updated when each worker _starts_, not finishes; but that should be fine since the workers run periodically anyways.

src/feeds/FeedReader.ts
```diff
@@ -189,7 +188,9 @@ export class FeedReader {
         log.debug('Loaded feed URLs:', this.observedFeedUrls);
 
         void this.loadSeenEntries().then(() => {
-            return this.pollFeeds();
+            for (let i = 0; i < config.pollConcurrency; i++) {
+                this.pollFeeds(i);
```
AndrewFerr, line 192: For consistency with line 406, which I hadn't noticed:
```suggestion
                void this.pollFeeds(i);
```
though its purely a stylistic choice (albeit one that makes it more clear that the returned Promises are meant to be ignored).

PR #795, opened 2023-07-06
Add more icons to GitHub messages

dav-is:
To make it easier to skim GitHub status updates, I think we should include an icon at the beginning of most messages.

I moved the approved/changes requested icon to the beginning of the message to be consistent and avoid breaking the flow of the words.

e.g. `User [checkmark icon] approved the following PR` seems cluttered compared to `[checkmark icon] User approved the following PR`.

Also, the length of the username would shift the position of the icon

Before:
```
John 📥 created an issue
Joe 📥 created an issue
Samantha 📥 created an issue
```

After:
```
📥 John created an issue
📥 Joe created an issue
📥 Samantha created an issue
```

### DIFF
--- a/changelog.d/795.feature
+++ b/changelog.d/795.feature
@@ -1,0 +1,1 @@
+Add more icons to GitHub repo hooks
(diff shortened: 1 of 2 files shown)


Half-Shot (approved):
These look great! If you can do a changelog entry we can merge this one.

PR #798, opened 2023-07-17
Remove Sled crypto store, use SQLite by default

AndrewFerr:
on account of it being removed from the crypto-sdk


docs/advanced/encryption.md
```diff
@@ -12,8 +12,7 @@ Hookshot supports end-to-bridge encryption via [MSC3202](https://github.com/matr
```
Half-Shot, line 12: I think this needs a changelog with a `.removal` and upgrade notes

Half-Shot (changes requested):
Looks good, mentioning the need for a changelog + upgrade notes.

changelog.d/798.removal
```diff
@@ -0,0 +1 @@
+Drop support for the Sled crypto store format. The configuration key of `experimentalEncryption.useLegacySledStore` is now ignored, and the crypto store will always use the SQLite format. If an existing SQLite store does not exist on bridge startup, one will be created.
```
Half-Shot, line 1: Sorry, another opinion. Can we have the app fail to start if this key is present to force users to take action?

AndrewFerr: Good idea

Half-Shot (approved):

PR #801, opened 2023-07-31
Make the Dockerfile cross-compile without qemu

sandhose:
This changes the Dockerfile so that it can be fully cross-compiled to arm64 & x86_64 without any qemu emulation.
This should mean way less time spent building the arm64 image in CI.
This also means better caching between the two architectures.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-napi = {version="2", features=["serde-json", "async"]}
+napi = { version = "2", features = ["serde-json", "async"] }
 napi-derive = "2"
 url = "2"
 serde_json = "1"
@@ -20,7 +20,7 @@ hex = "0.4.3"
 rss = "2.0.3"
 atom_syndication = "0.12"
 ruma = { version = "0.8.2", features = ["events", "unstable-sanitize"] }
-reqwest = "0.11"
+reqwest = { version = "0.11", default-features = false, features = ["rustls-tls-native-roots"] }
 
 [build-dependencies]
 napi-build = "1"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,51 +1,87 @@
-# Stage 0: Build the thing
-# Need debian based image to build the native rust module
-# as musl doesn't support cdylib
-FROM node:18 AS builder
+# syntax = docker/dockerfile:1.4
+
+# The Debian version and version name must be in sync
+ARG DEBIAN_VERSION=11
+ARG DEBIAN_VERSION_NAME=bullseye
+ARG RUSTC_VERSION=1.71.0
+ARG ZIG_VERSION=0.10.1
+ARG NODEJS_VERSION=18
+ARG CARGO_ZIGBUILD_VERSION=0.16.12
+# This needs to be kept in sync with the version in the package.json
+ARG MATRIX_SDK_VERSION=0.1.0-beta.6
+
+# Stage 1: Build the native rust module and the frontend assets
+FROM --platform=${BUILDPLATFORM} node:${NODEJS_VERSION}-${DEBIAN_VERSION_NAME} AS builder
+
+ARG CARGO_ZIGBUILD_VERSION
+ARG RUSTC_VERSION
+ARG ZIG_VERSION
+ARG TARGETPLATFORM
 
 # We need rustup so we have a sensible rust version, the version packed with bullsye is too old
-RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal
+RUN curl --proto '=https' --tlsv1.2 -sSf  https://sh.rustup.rs | sh -s -- -y --default-toolchain "${RUSTC_VERSION}" --profile minimal
 ENV PATH="/root/.cargo/bin:${PATH}"
 
-# arm64 builds consume a lot of memory if `CARGO_NET_GIT_FETCH_WITH_CLI` is not
-# set to true, so we expose it as a build-arg.
-ARG CARGO_NET_GIT_FETCH_WITH_CLI=false
-ENV CARGO_NET_GIT_FETCH_WITH_CLI=$CARGO_NET_GIT_FETCH_WITH_CLI
+RUN rustup target add  \
+  --toolchain "${RUSTC_VERSION}" \
+  x86_64-unknown-linux-gnu \
+  aarch64-unknown-linux-gnu
 
-# Needed to build rust things for matrix-sdk-crypto-nodejs
-# See https://github.com/matrix-org/matrix-rust-sdk-bindings/blob/main/crates/matrix-sdk-crypto-nodejs/release/Dockerfile.linux#L5-L6
-RUN apt-get update && apt-get install -y build-essential cmake
+# Install zig and cargo-zigbuild, which are used by napi to cross-compile the native module
+RUN \
+  curl --proto '=https' --tlsv1.2 -sSf "https://ziglang.org/download/${ZIG_VERSION}/zig-linux-$(uname -m)-${ZIG_VERSION}.tar.xz" | tar -J -x -C /usr/local && \
+  ln -s "/usr/local/zig-linux-$(uname -m)-${ZIG_VERSION}/zig" /usr/local/bin/zig
+
+RUN cargo install --locked cargo-zigbuild@=${CARGO_ZIGBUILD_VERSION}
 
 WORKDIR /src
 
 COPY package.json yarn.lock ./
 RUN yarn config set yarn-offline-mirror /cache/yarn
-RUN yarn --ignore-scripts --pure-lockfile --network-timeout 600000
-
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -20,7 +20,7 @@ hex = "0.4.3"
 rss = "2.0.3"
 atom_syndication = "0.12"
 ruma = { version = "0.8.2", features = ["events", "unstable-sanitize"] }
-reqwest = "0.11"
```
sandhose, line 23: I had to change that because the default features use native openssl bindings, which was harder to cross-compile

Dockerfile
```diff
@@ -1,51 +1,87 @@
...
-RUN yarn --network-timeout 600000 --production --pure-lockfile && yarn cache clean
+RUN --mount=type=cache,target=/cache/yarn \
+    sh ./docker-cross-env.sh "$TARGETPLATFORM" \
+    yarn --ignore-scripts --pure-lockfile --network-timeout 600000 --production
+# Workaround: the install script of the matrix-rust-sdk only installs for the current platform, not the target one
+RUN sh ./docker-download-sdk.sh "$TARGETPLATFORM" "$MATRIX_SDK_VERSION"
+
+
+# Stage 3: Build the final runtime image
+FROM --platform=$TARGETPLATFORM gcr.io/distroless/nodejs${NODEJS_VERSION}-debian${DEBIAN_VERSION}:nonroot AS runtime
```
sandhose, line 73: Please please note that this switches the end image to a distroless-based one, with none of the base tools. It should be lighter and more secure, but it also makes it that you can't pop a basic shell in the container anymore. I can roll back this change as needed, or consider using the `nonroot-debug` image, which includes `busybox` (so a light shell and common utilities)

Dockerfile
```diff
@@ -1,51 +1,87 @@
...
+
+COPY yarn.lock package.json scripts/docker-cross-env.sh scripts/docker-download-sdk.sh ./
 RUN yarn config set yarn-offline-mirror /cache/yarn
 
-RUN yarn --network-timeout 600000 --production --pure-lockfile && yarn cache clean
+RUN --mount=type=cache,target=/cache/yarn \
+    sh ./docker-cross-env.sh "$TARGETPLATFORM" \
+    yarn --ignore-scripts --pure-lockfile --network-timeout 600000 --production
+# Workaround: the install script of the matrix-rust-sdk only installs for the current platform, not the target one
+RUN sh ./docker-download-sdk.sh "$TARGETPLATFORM" "$MATRIX_SDK_VERSION"
```
sandhose, line 69: Did a PR for that: https://github.com/matrix-org/matrix-rust-sdk-crypto-nodejs/pull/8

Cargo.toml
Half-Shot: I doubt we mind all that much :+1:

Dockerfile
Half-Shot: We tend to prefer at least a base shell for inspection over lightness, how bad would it be to have the debug one?

Half-Shot (changes requested):
This looks great wow! Thanks so much for spending the time on it. I've added a few questions or thoughts where I had them, but I'd be happy to go with this.

Dockerfile
AndrewFerr: With that PR merged, would this line call `docker-cross-env.sh` instead?

scripts/docker-cross-env.sh
```diff
@@ -0,0 +1,31 @@
...
+  exit 1
+fi
+
+PLATFORM="$1"
+shift
+
+if [ "$PLATFORM" = "linux/arm64" ] || [ "$PLATFORM" = "linux/arm64/v8" ]; then
+  export npm_config_target_platform=linux
+  export npm_config_target_arch=arm64
+  export npm_config_target_libc=glibc
```
AndrewFerr, line 19: Are these variables the ones you mentioned in https://github.com/matrix-org/matrix-rust-sdk-crypto-nodejs/pull/8 as being understood by `node-gyp`?

Dockerfile
sandhose: The `yarn install` above would be enough if we remove the `--ignore-scripts` flag

Dockerfile
```diff
@@ -1,51 +1,87 @@
...
+FROM --platform=${BUILDPLATFORM} node:${NODEJS_VERSION}-${DEBIAN_VERSION_NAME} AS builder
+
+ARG CARGO_ZIGBUILD_VERSION
+ARG RUSTC_VERSION
+ARG ZIG_VERSION
+ARG TARGETPLATFORM
 
 # We need rustup so we have a sensible rust version, the version packed with bullsye is too old
-RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal
+RUN curl --proto '=https' --tlsv1.2 -sSf  https://sh.rustup.rs | sh -s -- -y --default-toolchain "${RUSTC_VERSION}" --profile minimal
```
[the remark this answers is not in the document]
sandhose: It's to ensure that there is no TLS downgrade attack 🤷 
It's the recommended flags you have on https://rustup.rs

Dockerfile
sandhose: So considering it's node anyway, so if you have an RCE, not having a shell will probably not stop you form doing anything anyway.

I see three options: 

 - provide both a debug and non-debug variant (which we basically get for free at this point) with two build targets. This is what I did on MAS: https://github.com/matrix-org/matrix-authentication-service/blob/d6130176b526f9e490ba6bfd2949694ddd6cc665/Dockerfile#L166-L186
 - flip to the debug variant if there is no value of having a non-debug one
 - go back to using `library/node` directly if we're not really concerned about lightness. It would also let us build in a bookworm-based environment, since the distroless images don't have Debian 12-based images yet
 
I really don't have strong opinions on any of those options 🤷

Dockerfile
```diff
@@ -1,51 +1,87 @@
-# Stage 0: Build the thing
-# Need debian based image to build the native rust module
-# as musl doesn't support cdylib
-FROM node:18 AS builder
+# syntax = docker/dockerfile:1.4
+
+# The Debian version and version name must be in sync
+ARG DEBIAN_VERSION=11
```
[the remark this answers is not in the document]
sandhose: See the other comment below, because distroless images don't support Debian 12 yet: https://github.com/GoogleContainerTools/distroless/issues/1337

Dockerfile
```diff
@@ -1,51 +1,87 @@
...
+ARG DEBIAN_VERSION_NAME=bullseye
+ARG RUSTC_VERSION=1.71.0
+ARG ZIG_VERSION=0.10.1
+ARG NODEJS_VERSION=18
+ARG CARGO_ZIGBUILD_VERSION=0.16.12
+# This needs to be kept in sync with the version in the package.json
+ARG MATRIX_SDK_VERSION=0.1.0-beta.6
```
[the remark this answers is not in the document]
sandhose: I mean, I expect that the next release will have the PR that avoids having to do that in the first place merged: https://github.com/matrix-org/matrix-rust-sdk-crypto-nodejs/pull/8

package.json
```diff
@@ -76,7 +76,7 @@
   },
   "devDependencies": {
     "@codemirror/lang-javascript": "^6.0.2",
-    "@napi-rs/cli": "^2.13.2",
+    "@napi-rs/cli": "3.0.0-alpha.5",
```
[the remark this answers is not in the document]
sandhose: I *think* I can avoid this though. You want me to try?

Dockerfile
sandhose: I ended up going back to library/node, which allowed me to:

 - pin the nodejs version better
 - make it based on bookworm
 - have a proper shell, avoid breaking anything in the image

package.json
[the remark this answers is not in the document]
sandhose: I did rollback to the current CLI tool version and it works 👍 

In the current CLI tool, support for zig is a bit more experimental (the new one relies on `cargo-zigbuild`, which is more reliable), so I had to rollback to zig 0.9.1 because I couldn't get it to work with 0.10.1

.dockerignore
```diff
@@ -7,6 +7,8 @@ public/
 registration.yml
 tests/
 tsconfig.tsbuildinfo
+Dockerfile
+.dockerignore
```
sandhose, line 11: This is to avoid voiding the cache every time I change the docker image, but doesn't change anything to the end image

Dockerfile
sandhose: I don't have write permissions on that repo so I can't merge myself
Besides, this would need a new release of the crypto SDK, which is probably not ready (especially since it moved repo)

package.json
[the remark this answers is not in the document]
Half-Shot: :pray:

Dockerfile
AndrewFerr: Ah, of course.  I'll get that kicked off then

PR #802, opened 2023-07-31
Add bridge port to Pod and Service manifests in Helm Chart

Hy3n4:
Hey there 👋 ,
I had this fight with the Hookshot installed via Helm Chart and this is the result. More info in matrix thread [here](https://matrix.to/#/!TlZdPIYrhwNvXlBiEk:half-shot.uk/$20_yOKi4pqE_VxpUsQOarGz9zwP3fjRblyCRSFgs2eM?via=half-shot.uk&via=matrix.org&via=element.io)

I hope I did not forgot anything and this PR is valid.


helm/hookshot/templates/_pod.tpl
```diff
@@ -48,6 +48,9 @@ containers:
         mountPath: "/data"
 {{- end }}
     ports:
+      - name: bridge
+        containerPort: {{ .Values.hookshot.config.bridge.port }}
```
Half-Shot, line 52: Heya, do you know what the bridge port corresponds to. This isn't the appservice (below)?

Hy3n4: Originally I thouth so, but I'm not sure, TBH. There is a port 9993 defined in the bridge configuration section. But in the service and container definitions are used hardcoded ports which does not correspond with bridge port. 🤷‍♂️

Hy3n4: So I've been digging more and the ports are really different. The `appservice` here is probably serving widget stuff, but the bridge itself is running on port defined in `.Values.hookshot.config.bridge.port`. Therefore, there must be another service for it.
Here is what ports are actually listening in the container:
image

And [Here](https://github.com/matrix-org/matrix-hookshot/blob/958743365f02b33dc6293ea226c1d6b8b14b0245/helm/hookshot/values.yaml#L64) in the `values.yaml`, there is `appservice` as listener.
So I think this PR is legit. 🤔 At least I couldn't make the hookshot work without these changes.

Hy3n4: Hey, It's been a while. Did you manage to get into this? I want to use the Helm chart installation, but I cannot, since the bridge port is not propagated on `Pod` or even available in the `Service`.

If I am doing something wrong, I would like to know what 🤷‍♂️ 
Thanks

rhizoet: Yes, can that please merged? We need to have Port 9993 opened in the cluster, so that we can reach it with the registration of synapse. Otherweise the helm chart ist useless.

PR #807, opened 2023-08-07
add dark mode and compound colors

HarHarLinks:
I've finally played around with hookshot dark mode and, while i'm at it, the new compound libraries.

`compound-design-tokens` seems to be where the colors are stored, at least that's the only place I found them.
<del>However, it puts them into css classes such as `.cpd-theme-light.cpd-theme-light` and I haven't found a way to access them based on media queries because of that. so right now I can only statically add the class e.g. `<html lang="en" class="cpd-theme-dark">` in index.html.</del>

<del>Best idea I got for now is adding the proper class onload with js based on `?theme=$org.matrix.msc2873.client_theme` that then I'd need to add to the widget URL.</del>

To do:
- [ ] ticky boxes to primary green
- [ ] use compound for green but disabled buttons
- [x] generic webhook and github icons are black pngs, i guess they would need to become svgs that use `--foreground-color`. this is kinda blocked on not having SVGs though

My dev env is local and I don't have access to all supported services, so my testing is limited, but here is how it looks for me:

light mode:

| before | after |
|---|---|
| image | image |
| image | image |
| image | image |

dark mode:

| before | after |
|---|---|
| image | image |
| image | image |
| image | image |
| note the icon | image |


web/App.tsx
```diff
@@ -130,6 +143,7 @@ export default class App extends Component<void, IState> {
                     supportedServices={this.state.supportedServices}
                     serviceScope={this.state.serviceScope}
                     embedType={this.state.embedType}
+                    darkMode={darkMode}
```
HarHarLinks, line 146: Passing darkmode as argument like this - while working - seems very explicit and noisy.

An alternative could be to inline SVGs into the DOM which would then allow the compound CSS to be applied to them. I would need to investigate how to do that. (We might not even want compound for this but our own media queries instead to us e.g. the official original GitHub colour, and then media queries are also not MSC2873 which we actually want etc etc).

Half-Shot:
So I've gone and added compound (which has both light and dark themes) to hookshot, so I think this issue is resolved. Thanks for kicking the work off though!!

PR #817, opened 2023-09-15
Replace vm2 with quickjs

Half-Shot:
`vm2` is no longer supported (https://github.com/patriksimek/vm2#%EF%B8%8F-project-discontinued-%EF%B8%8F). In the interests of keeping the experimental transformation function code safe and secure, we've going to try out quickjs within a wasm wrapper to ensure things are still safe to run.

This passes our basic sanity tests, but it will likely result in some breakages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,11 +65,11 @@
     "nyc": "^15.1.0",
     "p-queue": "^6.6.2",
     "prom-client": "^14.2.0",
+    "quickjs-emscripten": "^0.23.0",
     "reflect-metadata": "^0.1.13",
     "source-map-support": "^0.5.21",
     "string-argv": "^0.3.1",
     "tiny-typed-emitter": "^2.1.0",
-    "vm2": "^3.9.18",
     "winston": "^3.3.3",
     "xml2js": "^0.5.0",
     "yaml": "^2.2.2"
--- a/src/Connections/GenericHook.ts
+++ b/src/Connections/GenericHook.ts
@@ -321,9 +321,11 @@ export class GenericHookConnection extends BaseConnection implements IConnection
             const ctxResult = ctx.evalCode(`const data = ${JSON.stringify(data)};\n\n${this.transformationFunction}`);
     
             if (ctxResult.error){
-                throw Error("Transformation failed to run " + ctxResult.error);
+                const err = ctx.dump(ctxResult.error);
+                throw Error("Transformation failed to run " + err);
             } else {
-                result = ctx.dump(result.value);
+                const value = ctx.getProp(ctx.global, 'result');
+                result = ctx.dump(value);
             }
         } finally {
             ctx.dispose();
--- a/src/Connections/GenericHook.ts
+++ b/src/Connections/GenericHook.ts
@@ -2,7 +2,7 @@ import { Connection, IConnection, IConnectionState, InstantiateConnectionOpts, P
 import { Logger } from "matrix-appservice-bridge";
 import { MessageSenderClient } from "../MatrixSender"
 import markdownit from "markdown-it";
-import { VMScript as Script, NodeVM } from "vm2";
+import { QuickJSRuntime, QuickJSWASMModule, newQuickJSWASMModule } from "quickjs-emscripten";
 import { MatrixEvent } from "../MatrixEvent";
 import { Appservice, Intent, StateEvent } from "matrix-bot-sdk";
 import { ApiError, ErrCode } from "../api";
@@ -65,6 +65,11 @@ const SANITIZE_MAX_BREADTH = 50;
  */
 @Connection
 export class GenericHookConnection extends BaseConnection implements IConnection {
+    private static quickModule: QuickJSWASMModule;
+
+    public static async initaliseQuickJS() {
+        GenericHookConnection.quickModule = await newQuickJSWASMModule();
+    }
 
     /**
      * Ensures a JSON payload is compatible with Matrix JSON requirements, such
@@ -197,7 +202,8 @@ export class GenericHookConnection extends BaseConnection implements IConnection
         GenericHookConnection.LegacyCanonicalEventType,
     ];
 
-    private transformationFunction?: Script;
+    private transformationRuntime?: QuickJSRuntime;
+    private transformationFunction?: string;
     private cachedDisplayname?: string;
     constructor(
         roomId: string,
@@ -211,7 +217,9 @@ export class GenericHookConnection extends BaseConnection implements IConnection
     ) {
         super(roomId, stateKey, GenericHookConnection.CanonicalEventType);
         if (state.transformationFunction && config.allowJsTransformationFunctions) {
-            this.transformationFunction = new Script(state.transformationFunction);
+            this.transformationRuntime = GenericHookConnection.quickModule.newRuntime();
+            // TODO: Validate code.
+            this.transformationFunction = state.transformationFunction;
         }
     }
 
@@ -261,13 +269,15 @@ export class GenericHookConnection extends BaseConnection implements IConnection
     public async onStateUpdate(stateEv: MatrixEvent<unknown>) {
         const validatedConfig = GenericHookConnection.validateState(stateEv.content as Record<string, unknown>, this.config.allowJsTransformationFunctions || false);
         if (validatedConfig.transformationFunction) {
-            try {
-                this.transformationFunction = new Script(validatedConfig.transformationFunction);
-            } catch (ex) {
-                await this.messageClient.sendMatrixText(this.roomId, 'Could not compile transformation function:' + ex, "m.text", this.intent.userId);
-            }
-        } else {
-            this.transformationFunction = undefined;
+            this.transformationRuntime?.dispose();
+            this.transformationRuntime = GenericHookConnection.quickModule.newRuntime();
+            this.transformationFunction = validatedConfig.transformationFunction;
+            // const codeEvalResult = this.transformationRuntime.newContext().evalCode(validatedConfig.transformationFunction);
+            // if (codeEvalResult.error) {
+            //     this.transformationRuntime?.dispose();
+            //     this.transformationRuntime = undefined;
+            //     await this.messageClient.sendMatrixText(this.roomId, 'Could not compile transformation function:' + codeEvalResult.error, "m.text", this.intent.userId);
+            // }
         }
         this.state = validatedConfig;
     }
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -65,11 +65,11 @@
     "nyc": "^15.1.0",
     "p-queue": "^6.6.2",
     "prom-client": "^14.2.0",
+    "quickjs-emscripten": "^0.23.0",
```
Half-Shot, line 68: This works, but I am wondering if we should continue our Rust porting work and instead use something like https://crates.io/crates/boa_engine

Half-Shot: I've started an effort in hs/boajs.

AndrewFerr:
This should actually be ready now. @Half-Shot give it a look when you can (I can't request a review since this is technically your branch).

src/Connections/GenericHook.ts
```diff
@@ -107,27 +111,26 @@ export class GenericHookConnection extends BaseConnection implements IConnection
         return obj;
     }
 
-    static validateState(state: Record<string, unknown>, allowJsTransformationFunctions?: boolean): GenericHookConnectionState {
+    static validateState(state: Record<string, unknown>): Omit<GenericHookConnectionState, "hookId"> {
```
Half-Shot, line 114: This feels a bit weird, why not just make hookId optional in the state as it was originally?

src/Connections/GenericHook.ts
```diff
@@ -148,9 +150,8 @@ export class GenericHookConnection extends BaseConnection implements IConnection
 
         return new GenericHookConnection(
             roomId,
-            state,
+            event.content,
```
Half-Shot, line 153: We definitely want to return the *validated* state to the constructor here, not the random crap a user might have put in the event.

src/Connections/GenericHook.ts
```diff
@@ -162,14 +163,12 @@ export class GenericHookConnection extends BaseConnection implements IConnection
         if (!config.generic) {
             throw Error('Generic Webhooks are not configured');
         }
-        const hookId = randomUUID();
-        const validState = GenericHookConnection.validateState(data, config.generic.allowJsTransformationFunctions || false);
-        await GenericHookConnection.ensureRoomAccountData(roomId, intent, hookId, validState.name);
-        await intent.underlyingClient.sendStateEvent(roomId, this.CanonicalEventType, validState.name, validState);
-        const connection = new GenericHookConnection(roomId, validState, hookId, validState.name, messageClient, config.generic, as, intent);
+        const connection = new GenericHookConnection(roomId, data, randomUUID(), messageClient, config.generic, as, intent);
```
Half-Shot, line 166: We've stopped validating the state here?

src/Connections/GenericHook.ts
```diff
@@ -197,22 +196,29 @@ export class GenericHookConnection extends BaseConnection implements IConnection
...
-        if (state.transformationFunction && config.allowJsTransformationFunctions) {
-            this.transformationFunction = new Script(state.transformationFunction);
-        }
+        const validState = GenericHookConnection.validateState(stateContent);
+        super(roomId, validState.name, GenericHookConnection.CanonicalEventType);
+        this.state = {
+            ...validState,
+            hookId,
+        };
+        void this.validateTransformationFunction();
```
Half-Shot, line 217: The previous version didn't actually validate the script at this point (I believe new Script() wouldn't do a thing until it was run). As much as it might feel weird, I'd rather not break that behaviour for now.

Sending matrix messages in the constructor is a *big* no-no too.

src/Connections/GenericHook.ts
```diff
@@ -258,18 +264,41 @@ export class GenericHookConnection extends BaseConnection implements IConnection
...
-    public async onStateUpdate(stateEv: MatrixEvent<unknown>) {
-        const validatedConfig = GenericHookConnection.validateState(stateEv.content as Record<string, unknown>, this.config.allowJsTransformationFunctions || false);
-        if (validatedConfig.transformationFunction) {
-            try {
-                this.transformationFunction = new Script(validatedConfig.transformationFunction);
-            } catch (ex) {
-                await this.messageClient.sendMatrixText(this.roomId, 'Could not compile transformation function:' + ex, "m.text", this.intent.userId);
-            }
+    public onStateUpdate(stateEv: MatrixEvent<unknown>) {
+        return this.updateConfig(stateEv.content as Record<string, unknown>);
```
Half-Shot, line 268: It feels like updateConfig could actually be in here, this function isn't achieving much.

src/Connections/GenericHook.ts
```diff
@@ -258,18 +264,41 @@ export class GenericHookConnection extends BaseConnection implements IConnection
...
+
+    private validateTransformationFunction(): Promise<void> {
+        let result = Promise.resolve();
+        if (!GenericHookConnection.quickModule || !this.state.transformationFunction) {
+            return result;
+        }
+        this.hasValidTransformation = false;
+        const ctx = GenericHookConnection.quickModule.newContext();
+        const codeEvalResult = ctx.evalCode(`function f(data) {${this.state.transformationFunction}}`);
+        if (codeEvalResult.error) {
```
Half-Shot, line 287: I believe this a breaking change also, we never validated the function until the webhook ran before, and simply let the user know when there was an error.

I'm not saying this isn't a good way to go in the future, but I'd rather not break behaviours in this PR.

src/Connections/GenericHook.ts
```diff
@@ -258,18 +264,41 @@ export class GenericHookConnection extends BaseConnection implements IConnection
...
+            const message = "Could not compile transformation function:\n```" + JSON.stringify(ctx.dump(codeEvalResult.error)) + "```";
+            codeEvalResult.error.dispose();
+            result = this.intent.sendEvent(this.roomId, {
+                msgtype: "m.text",
+                body: message,
+                formatted_body: md.renderInline(message).replaceAll("\n", "<br>"),
```
Half-Shot, line 293: I'd hope that renderInline would aleady be using HTML elements...

src/Connections/GenericHook.ts
```diff
@@ -355,10 +393,10 @@ export class GenericHookConnection extends BaseConnection implements IConnection
      * @returns `true` if the webhook completed, or `false` if it failed to complete
      */
     public async onGenericHook(data: unknown): Promise<boolean> {
-        log.info(`onGenericHook ${this.roomId} ${this.hookId}`);
+        log.info(`onGenericHook ${this.roomId} ${this.state.hookId}`);
```
Half-Shot, line 396: What's the reason for changing this?

tests/connections/GenericHookTest.ts
```diff
@@ -175,7 +178,7 @@ describe("GenericHookConnection", () => {
     });
     it("will handle a hook event with a v2 transformation function", async () => {
         const webhookData = {question: 'What is the meaning of life?', answer: 42};
-        const [connection, mq] = createGenericHook({name: 'test', transformationFunction: V2TFFunction}, {
+        const [connection, mq] = createGenericHook({name: 'test', hookId: 'test', transformationFunction: V2TFFunction}, {
```
Half-Shot, line 181: I believe it's still technically valid to have these without hookIds?

Half-Shot (commented):
God I wish GitHub let you switch sides in reviews.

I'm minded to say that we want to be extremely careful when changing behaviours here. We're going to be asking people to upgrade ASAP as soon as this is ready, so we don't want to break anything we can avoid. If we can go back to failing on webhook with an error rather than pre-validating, that would be good.

src/Connections/GenericHook.ts
AndrewFerr: This was part of making `hookId` a mandatory property, lest it's not set in some state update/creation functions.  Not sure how important that is, so I can revert that if needed.

src/Connections/GenericHook.ts
AndrewFerr: The constructor now does the validating, so that callers don't have to be mindful to pre-validate it.  But it looks like that should be reverted.

src/Connections/GenericHook.ts
AndrewFerr: Ditto my previous comment

src/Connections/GenericHook.ts
AndrewFerr: Ok, I'll revert the validation in the constructor, but will at least add a docstring to say that the `state` argument ought to come from a call to `validateState`.

src/Connections/GenericHook.ts
AndrewFerr: This is here to wrap around `updateConfig`, which is also called by `provisionerUpdateConfig`.  i.e. a config update can come from either a direct update to its room state event, or from a call to the provisioning API; this light `onStateUpdate` is just a thin layer to make `updateConfig` compatible with `IConnectionState#onStateUpdate`.

src/Connections/GenericHook.ts
AndrewFerr: It got validated on a state event update, and using the provisioning API would both call the constructor & send a state event which Hookshot would see as an update.  So in practice, it did get validated before being run, except in some tests that only call the constructor without sending a state event.

src/Connections/GenericHook.ts
AndrewFerr: I have a commit that fixes this.  Luckily the answer was just a few lines below :slightly_smiling_face:

src/Connections/GenericHook.ts
AndrewFerr: It was to deduplicate the `hookId` from being both a property of `GenericHookConnection` and its `state` object, since in practice the `hookId` is always set to the `state`'s `name`.

tests/connections/GenericHookTest.ts
AndrewFerr: This was required due to having made `hookId` mandatory, but that can be reverted.

src/Connections/GenericHook.ts
AndrewFerr: Reverted

src/Connections/GenericHook.ts
AndrewFerr: Reverted

src/Connections/GenericHook.ts
AndrewFerr: Validation reverted and docstring added.

src/Connections/GenericHook.ts
AndrewFerr: Reverted, though it retains the inconsistency of how the transformation function is validated & set between state updates & the constructor.  We can revisit that in a later change, though.

src/Connections/GenericHook.ts
AndrewFerr: Fixed, it now looks like this:
image

src/Connections/GenericHook.ts
```diff
@@ -301,20 +330,29 @@ export class GenericHookConnection extends BaseConnection implements IConnection
...
     public executeTransformationFunction(data: unknown): {plain: string, html?: string, msgtype?: string}|null {
-        if (!this.transformationFunction) {
-            throw Error('Transformation function not defined');
-        }
-        const vm = new NodeVM({
-            console: 'off',
-            wrapper: 'none',
-            wasm: false,
-            eval: false,
-            timeout: TRANSFORMATION_TIMEOUT_MS,
```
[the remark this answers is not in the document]
AndrewFerr: 31dcf7c

tests/connections/GenericHookTest.ts
AndrewFerr: Reverted

src/Connections/GenericHook.ts
AndrewFerr: Reverted

src/Connections/GenericHook.ts
AndrewFerr: Reverted

Half-Shot:
:heavy_check_mark:

PR #818, opened 2023-09-26
Fix top level return incompatibility

Half-Shot:
Missed in #817

### DIFF
--- a/src/Connections/GenericHook.ts
+++ b/src/Connections/GenericHook.ts
@@ -332,7 +332,7 @@ export class GenericHookConnection extends BaseConnection implements IConnection
         ctx.runtime.setInterruptHandler(shouldInterruptAfterDeadline(Date.now() + TRANSFORMATION_TIMEOUT_MS));
         try {
             ctx.setProp(ctx.global, 'HookshotApiVersion', ctx.newString('v2'));
-            const ctxResult = ctx.evalCode(`const data = ${JSON.stringify(data)};\n\n${this.state.transformationFunction}`);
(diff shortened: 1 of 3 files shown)


src/Connections/GenericHook.ts
```diff
@@ -332,7 +332,7 @@ export class GenericHookConnection extends BaseConnection implements IConnection
         ctx.runtime.setInterruptHandler(shouldInterruptAfterDeadline(Date.now() + TRANSFORMATION_TIMEOUT_MS));
         try {
             ctx.setProp(ctx.global, 'HookshotApiVersion', ctx.newString('v2'));
-            const ctxResult = ctx.evalCode(`const data = ${JSON.stringify(data)};\n\n${this.state.transformationFunction}`);
+            const ctxResult = ctx.evalCode(`const data = ${JSON.stringify(data)};\n(() => { ${this.state.transformationFunction} })();`);
```
Half-Shot, line 335: Obviously users can break out of this function, but I don't think that makes much difference.

tests/connections/GenericHookTest.ts
```diff
@@ -199,6 +200,29 @@ describe("GenericHookConnection", () => {
             type: 'm.room.message',
         });
     });
+    it("will handle a hook event with a top-level return", async () => {
```
Half-Shot, line 203: Confirmed this breaks on `main` and works in this branch.

AndrewFerr (approved):

src/Connections/GenericHook.ts
AndrewFerr: True, it shouldn't make a difference because nothing comes after the arrow function, so nothing would happen after the return.

PR #819, opened 2023-09-26
Update release script

AndrewFerr:
This will help prevent cutting releases where the Cargo package version wasn't updated to match what's in package.json.


Half-Shot (approved):
Really cool, thank you!

Issue #821, opened 2023-10-06
Helm chart enables queue.host,port configuration by default, which causes it to talk to a nonexistant Redis


Half-Shot:
We should disable this by default / add a Redis to the helm.

laurencegill:
IIRC `github`, `gitlab`, `figma` and `jira` all need to be optional too

Half-Shot:
Fixed in #823

PR #823, opened 2023-10-09
Hookshot optionals

laurencegill:
Fix for https://github.com/matrix-org/matrix-hookshot/issues/821 and https://github.com/vector-im/sre-internal/issues/2840

### DIFF
--- a/helm/hookshot/values.yaml
+++ b/helm/hookshot/values.yaml
@@ -243,35 +243,35 @@ hookshot:
       colorize: true
       json: false
       timestampFormat: HH:mm:ss:SSS
-    widgets:
+    # widgets:
       # (Optional) EXPERIMENTAL support for complimentary widgets
       #
-      addToAdminRooms: false
-      disallowedIpRanges:
-        - 127.0.0.0/8
-        - 10.0.0.0/8
-        - 172.16.0.0/12
-        - 192.168.0.0/16
-        - 100.64.0.0/10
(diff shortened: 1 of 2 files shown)


helm/hookshot/values.yaml
```diff
@@ -129,64 +129,64 @@ hookshot:
     github:
```
Half-Shot, line 129: I think also comment out these top level things too

Half-Shot (changes requested):
One thought, otherwise thank you :)

helm/hookshot/values.yaml
laurencegill: Happy with that, although testing showed it seems to work with them defined but not :smile: 

What about [widgets](https://github.com/matrix-org/matrix-hookshot/blob/main/helm/hookshot/values.yaml#L246) while we're at it?

Ideally, but no showstopped, AFAIU its best practice to use an empty mapping such as:
```
    queue: {}
```

However I tested this and it caused the same behaviour (looking for localhost) so I suspect it's something in the code that falls back to that?

Half-Shot: Queue should *just* contain `monolithic: true`, rather than empty object.

laurencegill: > Queue should _just_ contain `monolithic: true`, rather than empty object.

As discussed we're leaving this commented out completely.

PR #826, opened 2023-10-10
Ensure sample config only defaults to bare minimum requirements

Half-Shot:
The previous version of the config included all the items as enabled, which broke bits of helm. Users should choose which features to enable instead.


laurencegill (approved):

laurencegill (approved):

Half-Shot:
(This CI is going to take ages and we know it passes)

PR #830, opened 2023-10-17
Add note about scope for private vs. public repo notifications

Twi1ightSparkle:
In hindsight it's very obvious that it need `repo` to get issue comments and other details for notifications from private repos, but it took me way to much time to think of that.

And, `repo:status` is above `public_repo` on the list of token permissions.


Half-Shot (approved):
Thanks!

PR #833, opened 2023-10-23
feat: add allowedIpRanges widgets config

megheaiulian:
Adds `allowedIpRanges` widgets config that can be used specifically
allow  some ip ranges even if its in `disallowedIpRanges`.


changelog.d/833.feature
```diff
@@ -0,0 +1 @@
+Adds `allowedIpRanges` widgets config
```
Half-Shot, line 1: I realise this might be slightly verbose, but just wanted to make it clear what this does.

```suggestion
Adds widgets config (`allowedIpRanges`) option to control which IPs are allowed to access widgets.
```

Half-Shot (approved):
Seems reasonable, thank you!

src/config/Config.ts
```diff
@@ -366,10 +369,14 @@ export class BridgeWidgetConfig {
     constructor(yaml: BridgeWidgetConfigYAML) {
         this.addToAdminRooms = yaml.addToAdminRooms || false;
         this.disallowedIpRanges = yaml.disallowedIpRanges;
+        this.allowedIpRanges = yaml.allowedIpRanges;
         this.roomSetupWidget = yaml.roomSetupWidget;
         if (yaml.disallowedIpRanges !== undefined && (!Array.isArray(yaml.disallowedIpRanges) || !yaml.disallowedIpRanges.every(s => typeof s === "string"))) {
             throw new ConfigError("widgets.disallowedIpRanges", "must be a string array");
         }
+        if (yaml.allowedIpRanges !== undefined && (!Array.isArray(yaml.allowedIpRanges) || !yaml.allowedIpRanges.every(s => typeof s === "string"))) {
```
Half-Shot, line 377: N.B. An empty array would skip this, would we allow empty values for allowedIpRanges (effectively blocking the feature entirely, seems a bit pointless :) )

src/config/Defaults.ts
```diff
@@ -40,6 +40,7 @@ export const DefaultConfigRoot: BridgeConfigRoot = {
         roomSetupWidget: {
             addOnInvite: false,
         },
+        allowedIpRanges: [],
```
Half-Shot, line 43: Wondering if we want this to actually show up as a default, because the default value would block everything currently :)

docs/advanced/widgets.md
```diff
@@ -39,6 +39,7 @@ widgets:
 #     - 2001:db8::/32
 #     - ff00::/8
 #     - fec0::/10
+# allowedIpRanges: []
```
Half-Shot, line 42: Ideally we would put in some examples of values here?

Half-Shot (changes requested):
A few things to think about. Generally looks good though :)

PR #841, opened 2023-11-11
Changed passFile value

mike-pisman:
Changed default location of the passkey.pem file and added useful comments on how to fill in passkey value

I think this makes it clearer...


Aldekein:
I think this commit should be merged there.

PR #846, opened 2023-11-22
Update releaser script to check staged contents

AndrewFerr:
to ensure that the package versions to be committed are correct.


Half-Shot (approved):
Good plan!

PR #876, opened 2024-01-03
Add command to list feeds in JSON and YAML format

Twi1ightSparkle:
The only other way currently to export all feeds from a room without loosing info is manually parsing the room state.

Screenshot 2024-01-04 at 11 08 41 AM

Screenshot 2024-01-04 at 11 09 03 AM

### DIFF
--- a/changelog.d/876.feature
+++ b/changelog.d/876.feature
@@ -1,0 +1,1 @@
+Add `webhook list-yaml` command to easily export all feeds from a room.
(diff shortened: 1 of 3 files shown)


Twi1ightSparkle:
I don't think my lint is bad, it failed on yarn --ignore-scripts 😕

Half-Shot (approved):
Super cool feature, this works for me :)

PR #890, opened 2024-02-06
Backoff failing RSS feeds

Half-Shot:
This should generally help speed up working feeds, by backing off failing ones. This is useful for environments with 1000s of feeds, some of which may have rotted since they were set up. For instance, integrations.ems.host sees about a 5% failure rate at the HTTP level alone.

Leftovers from this PR:
 - Cache backoff times across restarts in Redis. We currently hit all feeds on startup and backoff again. We would need to provide a flush button though.
 - General CPU improvements, now that we're failing less, the CPU is working harder.

### DIFF
--- a/src/Stores/RedisStorageProvider.ts
+++ b/src/Stores/RedisStorageProvider.ts
@@ -29,11 +29,10 @@ const WIDGET_USER_TOKENS = "widgets.user-tokens.";
 
 const FEED_GUIDS = "feeds.guids.";
 
-
-
 const log = new Logger("RedisASProvider");
 
 export class RedisStorageContextualProvider implements IStorageProvider {
+
     constructor(protected readonly redis: Redis, protected readonly contextSuffix = '') { }
 
     public setSyncToken(token: string|null){
@@ -216,9 +215,9 @@ export class RedisStorageProvider extends RedisStorageContextualProvider impleme
         await this.redis.set(key, JSON.stringify(value));
     }
 
-    public async storeFeedGuids(url: string, ...guid: string[]): Promise<void> {
+    public async storeFeedGuids(url: string, ...guids: string[]): Promise<void> {
         const feedKey = `${FEED_GUIDS}${url}`;
-        await this.redis.lpush(feedKey, ...guid);
+        await this.redis.lpush(feedKey, ...guids);
         await this.redis.ltrim(feedKey, 0, MAX_FEED_ITEMS);
     }
 
@@ -226,7 +225,16 @@ export class RedisStorageProvider extends RedisStorageContextualProvider impleme
         return (await this.redis.exists(`${FEED_GUIDS}${url}`)) === 1;
     }
 
-    public async hasSeenFeedGuid(url: string, guid: string): Promise<boolean> {
-        return (await this.redis.lpos(`${FEED_GUIDS}${url}`, guid)) != null;
+    public async hasSeenFeedGuids(url: string, ...guids: string[]): Promise<string[]> {
+        let multi = this.redis.multi();
+        for (const guid of guids) {
+            multi = multi.lpos(`${FEED_GUIDS}${url}`, guid);
+        }
+        const res = await multi.exec();
+        if (res === null) {
+            // Just assume we've seen none.
+            return [];
+        }
+        return guids.filter((_guid, index) => res[index][1] !== null);
     }
 }
--- a/src/Stores/StorageProvider.ts
+++ b/src/Stores/StorageProvider.ts
@@ -25,7 +25,7 @@ export interface IBridgeStorageProvider extends IAppserviceStorageProvider, ISto
     setStoredTempFile(key: string, value: string): Promise<void>;
     getGitlabDiscussionThreads(connectionId: string): Promise<SerializedGitlabDiscussionThreads>;
     setGitlabDiscussionThreads(connectionId: string, value: SerializedGitlabDiscussionThreads): Promise<void>;
-    storeFeedGuids(url: string, ...guid: string[]): Promise<void>;
-    hasSeenFeed(url: string, ...guid: string[]): Promise<boolean>;
-    hasSeenFeedGuid(url: string, guid: string): Promise<boolean>;
+    storeFeedGuids(url: string, ...guids: string[]): Promise<void>;
+    hasSeenFeed(url: string): Promise<boolean>;
+    hasSeenFeedGuids(url: string, ...guids: string[]): Promise<string[]>;
 }
--- a/src/feeds/parser.rs
+++ b/src/feeds/parser.rs
@@ -70,7 +70,8 @@ fn parse_channel_to_js_result(channel: &Channel) -> JsRssChannel {
                     .map(|f| f.value)
                     .or(item.link.clone())
                     .or(item.title.clone())
-                    .and_then(|f| Some(format!("md5:{:?}", hash_id(f)))),
+                    .and_then(|f| hash_id(f).ok())
+                    .and_then(|f| Some(format!("md5:{}", f)))
             })
             .collect(),
     }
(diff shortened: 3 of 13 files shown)


src/util/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        for item in self.backoff.iter() {
+            if *item.1 < since_the_epoch.as_millis() {
+                self.queue.push_back(item.0.clone());
+                items_to_rm.push(item.0.clone());
+            }
+        }
+
+        for item in items_to_rm {
+            self.backoff.remove(&item);
+        }
```
Half-Shot, line 50: Something about this strikes me as inefficient. I sort of want to effectively pop only some items from the HashMap. Currently we have to do a O(N) iteration on each pop, which is pretty rubbish.

src/Stores/StorageProvider.ts
```diff
@@ -25,7 +25,7 @@ export interface IBridgeStorageProvider extends IAppserviceStorageProvider, ISto
     setStoredTempFile(key: string, value: string): Promise<void>;
     getGitlabDiscussionThreads(connectionId: string): Promise<SerializedGitlabDiscussionThreads>;
     setGitlabDiscussionThreads(connectionId: string, value: SerializedGitlabDiscussionThreads): Promise<void>;
-    storeFeedGuids(url: string, ...guid: string[]): Promise<void>;
-    hasSeenFeed(url: string, ...guid: string[]): Promise<boolean>;
-    hasSeenFeedGuid(url: string, guid: string): Promise<boolean>;
+    storeFeedGuids(url: string, ...guids: string[]): Promise<void>;
+    hasSeenFeed(url: string): Promise<boolean>;
+    hasSeenFeedGuids(url: string, ...guids: string[]): Promise<string[]>;
```
Half-Shot, line 30: Sped this up by checking multiple guids.

src/Stores/RedisStorageProvider.ts
```diff
@@ -216,17 +215,26 @@ export class RedisStorageProvider extends RedisStorageContextualProvider impleme
...
     }
 
-    public async hasSeenFeedGuid(url: string, guid: string): Promise<boolean> {
-        return (await this.redis.lpos(`${FEED_GUIDS}${url}`, guid)) != null;
+    public async hasSeenFeedGuids(url: string, ...guids: string[]): Promise<string[]> {
+        let multi = this.redis.multi();
+        for (const guid of guids) {
+            multi = multi.lpos(`${FEED_GUIDS}${url}`, guid);
+        }
+        const res = await multi.exec();
```
Half-Shot, line 233: There was a CPU hit from doing a lot of hits to redis at once. This at least avoids multiple transactions, and at least gives Redis the option of optimising many `lpos` requests. 
Annoyingly it doesn't have a way to ask for multiple list entries at once though.

src/feeds/parser.rs
```diff
@@ -70,7 +70,8 @@ fn parse_channel_to_js_result(channel: &Channel) -> JsRssChannel {
                     .map(|f| f.value)
                     .or(item.link.clone())
                     .or(item.title.clone())
-                    .and_then(|f| hash_id(f).ok()),
+                    .and_then(|f| hash_id(f).ok())
+                    .map(|f| format!("md5:{}", f)),
```
Half-Shot, line 74: Cheap optimisation to do this Rust-side.

src/util/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+        let since_the_epoch = start.duration_since(UNIX_EPOCH).unwrap().as_millis() as u64;
+
+        // We only need to check this once, as we won't be adding to the backoff queue
+        // as often as we pull from it.
+        if let Some(item) = self.backoff.first_entry() {
+            if *item.key() < since_the_epoch {
+                let v = item.remove();
+                self.queue.push_back(v);
+            }
+        }
```
AndrewFerr, line 64: Could you explain this part?  I'm reading it as "if the lowest backoff duration has passed, enqueue the URL that the backoff was for".  If that's the case, why is it safe to check that only once?  Is it because the queue is only for items whose backoff has expired & are thus always fair game to be fetched -- and because if no item has finished its backoff, this will return `None` and the caller will see no feeds available to poll?

src/util/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+
+        self.queue.pop_front()
+    }
+
+    #[napi]
+    pub fn remove(&mut self, item: String) -> bool {
+        // Remove from the queue
+        if let Ok(index) = self.queue.binary_search(&item) {
+            self.queue.remove(index);
+            return true;
```
AndrewFerr, line 74: Nitpick: be rusty
```suggestion
            true
```

src/util/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+
+        self.queue.pop_front()
+    }
+
+    #[napi]
+    pub fn remove(&mut self, item: String) -> bool {
+        // Remove from the queue
+        if let Ok(index) = self.queue.binary_search(&item) {
```
AndrewFerr, line 72: An item can be in either the queue or the backoff map, but not both, right?

src/util/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+                    found_key = *key;
+                }
+            }
+            if found_key != 0 {
+                self.backoff.remove(&found_key);
+                return true;
+            }
+        }
+        // Always remove the duration on removal.
+        self.last_backoff_duration.remove(&item);
```
AndrewFerr, line 91: When would an item without a backoff have an entry in this?
Also, should this removal be done even when the entry does have a backoff?

src/util/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+        backoff_duration
+    }
+
+    #[napi]
+    pub fn length(&self) -> u32 {
+        self.queue.len() as u32
+    }
+
+    #[napi]
+    pub fn shuffle(&mut self) {
```
AndrewFerr, line 136: This could be made private

src/util/mod.rs
Half-Shot: Sure, and please check my math here. We check the backoff queue once per `pull` call, which means we will pull one item (if available) and push it onto the "active" queue.

Each time we pull, we will do an action and then either push onto the active queue or the backoff queue. This means at worst case, we are only ever putting onto the backoff queue as often as we are pulling from it. In reality, it's unlikely that you are backing off every feed but if you were, you'd essentially just have a queue with a delay in it. The other thing to note here is that backed off items are inherently delayed, so that it's reasonably possible that we are `pull`ing more often than the items become "ready".

Now, it is possible that several items could all become ready within a few `ms` and the pull code isn't called exactly when they become ready. That's not really a problem, because we don't mind if items that are already failing are waiting a bit longer -- as long as they eventually get polled, they will enter the fast lane again. 

So to summarise why this is safe, we won't ever push to the backoff queue more often than we pull.

This is also efficient, because we're doing less work per pull (not iterating) and end users should notice no difference.

Does that make more sense?

src/util/mod.rs
Half-Shot: yes

src/util/mod.rs
Half-Shot: So if the item has never backed off, this is empty. If a entry that had previously failed had started working again, then we remove it. This map is purely to track that if we repeatedly back off an item, that we actually do increase the timeout rather than just timing out for the same duration each time.

src/util/mod.rs
Half-Shot: apparently this was not allowed, maybe because it's not the last line in the function?

src/util/mod.rs
AndrewFerr: Indeed it does, thanks!

AndrewFerr (approved):

PR #895, opened 2024-02-13
Improve request throughput

Half-Shot:
Larger deployments seem to struggle with many requests against the same host (i.e. the homeserver), so this change replaces the existing request function with a `fetch`-powered function. This also enables support for HTTP/2 / pipelining so in theory we should go a bit faster than at present.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2616,11 +2616,11 @@ axios@^0.27.2:
     form-data "^4.0.0"
 
 axios@^1.6.3:
-  version "1.6.3"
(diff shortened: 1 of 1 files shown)


Half-Shot:
Testing this showed that the H2 client for `fetch` / `undici` was incomplete / buggy and we probably need to hold off for now.

Half-Shot:
https://github.com/nodejs/undici/issues/2704 was biting us.

Half-Shot:
Tested again on staging and it seems to work, but holding off until undici release the fix for #2704

PR #899, opened 2024-02-21
Change example Openid Connect callback url's to be correct.

minecraftchest1:
The docs suggested that the Openid Connect callback url's should be at `/bridge_oauth`, but it should actually point to `/oauth

Resolves https://matrix.to/#/!TlZdPIYrhwNvXlBiEk:half-shot.uk/$surQW-4XHEJ8wgzzOClOF0QoWM-jjA1V5X3BlNQfQaw?via=half-shot.uk&via=matrix.org&via=element.io`

### DIFF
--- a/config.sample.yml
+++ b/config.sample.yml
@@ -50,7 +50,7 @@ listeners:
 #    # (Optional) Settings for allowing users to sign in via OAuth.
 #    client_id: foo
 #    client_secret: bar
-#    redirect_uri: https://example.com/bridge_oauth/
+#    redirect_uri: https://example.com/oauth/
 #  defaultOptions:
 #    # (Optional) Default options for GitHub connections.
 #    showIssueRoomLink: false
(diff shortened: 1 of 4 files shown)


config.sample.yml
```diff
@@ -50,7 +50,7 @@ listeners:
 #    # (Optional) Settings for allowing users to sign in via OAuth.
 #    client_id: foo
 #    client_secret: bar
-#    redirect_uri: https://example.com/bridge_oauth/
+#    redirect_uri: https://example.com/oauth/
```
Half-Shot, line 53: You'll need to update it in https://github.com/matrix-org/matrix-hookshot/blob/main/src/config/Defaults.ts#L69 too, as this file is generated from that.

Half-Shot (changes requested):
tweak the config .ts files, and add a newsfile and you're done!

minecraftchest1:
Should I create two changelog files (one for the defaults in the source code, and one for the docs)?

Half-Shot (approved):
Nice!

PR #902, opened 2024-02-26
Split cache config from queue config.

Half-Shot:
The reason these were conjoined is because we originally wanted a Redis for worker mode, and eventually started using it as a cache. These days we recommend using Redis as a cache, but not for message queueing. This is really confusing, and to be honest could easily be seperated into two configs.

This changes does just that. We're also going to move towards `redisUri` because frankly it allows more options for less work.

### DIFF
--- a/src/config/Config.ts
+++ b/src/config/Config.ts
@@ -10,6 +10,8 @@ import { ConfigError } from "../errors";
 import { ApiError, ErrCode } from "../api";
 import { GithubInstance, GITHUB_CLOUD_URL } from "../github/GithubInstance";
 import { Logger } from "matrix-appservice-bridge";
+import { BridgeConfigCache } from "./sections/cache";
+import { BridgeConfigQueue } from "./sections";
 
 const log = new Logger("Config");
 
@@ -407,12 +409,6 @@ interface BridgeConfigWebhook {
     bindAddress?: string;
 }
 
-export interface BridgeConfigQueue {
-    monolithic: boolean;
-    port?: number;
-    host?: string;
-}
-
 export interface BridgeConfigLogging {
     level: "debug"|"info"|"warn"|"error"|"trace";
     json?: boolean;
@@ -454,6 +450,7 @@ export interface BridgeConfigSentry {
     environment?: string;
 }
 
+
 export interface BridgeConfigRoot {
     bot?: BridgeConfigBot;
     serviceBots?: BridgeConfigServiceBot[];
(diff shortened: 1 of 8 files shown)


docs/advanced/workers.md
```diff
@@ -11,18 +11,18 @@ This feature is <b>experimental</b> and should only be used when you are reachin
...
 
 The processes should all share the same config, which should contain the correct config to enable Redis:
 
 ```yaml
 queue:
-  monolithic: false
-  port: 6379
-  host: github-bridge-redis
+  redisUri: "redis://redis-host:6379"
+cache:
```
HarHarLinks, line 21: Is this needed here? As it's already mentioned below.

Half-Shot: I'm going to keep it out of caution, as it could look like `queue` replaces `cache` having the same properties. Specifying it twice does no harm either.

config.sample.yml
```diff
@@ -134,12 +134,15 @@ listeners:
...
 
+#cache:
+#  # (Optional) Cache options for large scale deployments.
+#  # For encryption to work, this must be configured.
+#  redisUri: redis://localhost:6379
+
 #queue:
-#  # (Optional) Message queue / cache configuration options for large scale deployments.
+#  # (Optional) Message queue configuration options for large scale deployments.
 #  # For encryption to work, must be set to monolithic mode and have a host & port specified.
```
AndrewFerr, line 144:
```suggestion
#  # For encryption to work, this must not be configured.
```

docs/setup.md
```diff
@@ -222,6 +222,20 @@ Please note that the appservice HTTP listener is configured <strong>separately</
...
+afford to, as it will generally improve startup times. Some features such as resuming RSS/Atom feeds between restarts
+is also only possible with a external cache.
+
+To enable, simply set:
+
+```yaml
+cache:
+  redisUri: "redis://redis-host:3679"
+```
+
```
AndrewFerr, line 237: docs/advanced/encryption.md needs updating to mention that crypto requires this config to be enabled (instead of `queue.monolithic`).

AndrewFerr (changes requested):
Just some documentation requests

src/config/Config.ts
```diff
@@ -454,40 +450,45 @@ export interface BridgeConfigSentry {
...
+ For encryption to work, this must be configured.`, true)
+    public readonly cache?: BridgeConfigCache;
     @configKey(`Configuration for encryption support in the bridge.
  If omitted, encryption support will be disabled.
  This feature is HIGHLY EXPERIMENTAL AND SUBJECT TO CHANGE.
  For more details, see https://github.com/matrix-org/matrix-hookshot/issues/594.`, true)
     public readonly encryption?: BridgeConfigEncryption;
-    @configKey(`Message queue / cache configuration options for large scale deployments.
+    @configKey(`Message queue configuration options for large scale deployments.
  For encryption to work, must be set to monolithic mode and have a host & port specified.`, true)
```
AndrewFerr, line 490:
```suggestion
 For encryption to work, this must not be configured.
```

AndrewFerr (changes requested):
The config in `helm/hookshot/values.yaml` should be updated too.

Half-Shot:
Believe everything has been addressed :partying_face:

PR #920, opened 2024-04-03
Add support for crypto store PVC to helm chart

PseudoResonance:
Adds a persistence section to the Helm chart so that the encryption crypto store can persist through restarts/updates.

### DIFF
--- a/helm/hookshot/values.yaml
+++ b/helm/hookshot/values.yaml
@@ -92,6 +92,15 @@ resources: {}
 #   cpu: 100m
 #   memory: 128Mi
 
+# Persistent volume for encryption store, mounted in the pod as /persistent.
+# Can be disabled if encryption is not used.
+persistence:
+  enabled: true
+  # existingClaim: hookshot
+  # storageClass: "-"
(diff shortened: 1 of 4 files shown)


helm/hookshot/values.yaml
```diff
@@ -92,6 +92,15 @@ resources: {}
 #   cpu: 100m
 #   memory: 128Mi
 
+# Persistent volume for encryption store, mounted in the pod as /persistent.
```
Half-Shot, line 95: Do we need to configure hookshot to use this?

Half-Shot (changes requested):
Looks okay, but one question.

PseudoResonance:
Yes, the following needs to be added to the Helm values under `hookshot.config`.
```yaml
    experimentalEncryption:
      storagePath: /persistent
```

rhizoet:
Can this please be merged? We need this feature very urgently, as there are currently only functioning webhooks with hookshot. And it is important that these are encrypted.

Thank you in advance.

PR #924, opened 2024-04-08
Add support for Challenge Hound

Half-Shot:
https://github.com/Half-Shot/matrix-challenger/ was written a long time ago before hookshot really supported multiple connections very well. The aim of this PR is to basically pull what we can from that project and insert it as a new connection type. The reader has some overlap with the RSS functionality, so maybe we should at some point refactor it all into a generic item polling class.


src/Connections/HoundConnection.ts
```diff
@@ -0,0 +1,167 @@
...
+        HoundConnection.LegacyEventType,
+    ];
+    static readonly ServiceCategory = "challengehound";
+
+
+    public static validateState(data: Record<string, unknown>): HoundConnectionState {
+        if (!data.url || typeof data.url !== "string") {
+            throw Error('Missing or invalid url');
+        }
+        const url = new URL(data.url);
```
Half-Shot, line 94: TODO: Validate that this is vaguely a correct URL

PR #926, opened 2024-04-16
Apply CSPs to generic hook responses.

Half-Shot:
A generic hook could potentially return a valid HTML document provided both `enableHttpGet` and JS functions are enabled. This isn't bad in itself but being able to execute JS is not really a useful feature.

### DIFF
--- a/src/generic/Router.ts
+++ b/src/generic/Router.ts
@@ -4,6 +4,7 @@ import { Logger } from "matrix-appservice-bridge";
 import { ApiError, ErrCode } from "../api";
 import { GenericWebhookEvent, GenericWebhookEventResult } from "./types";
 import * as xml from "xml2js";
+import helmet, { crossOriginOpenerPolicy } from "helmet";
 
 const WEBHOOK_RESPONSE_TIMEOUT = 5000;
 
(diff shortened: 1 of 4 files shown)


src/generic/Router.ts
```diff
@@ -83,6 +84,17 @@ export class GenericWebhooksRouter {
         const router = Router();
         router.all(
             '/:hookId',
+            helmet({
+                contentSecurityPolicy: {
+                    useDefaults: true,
+                    directives: {
+                        default: 'self',
```
davidegirardi, line 91: This  renders a CSP containing `defautl self;` which is not valid AFAIK. Did you mean to spell out `default-src`?
It's set to `'self'` in the helmet default but I see why you want to avoid unexpected changes.

```suggestion
                        defaultSrc: "'self'",
```

src/generic/Router.ts
```diff
@@ -83,6 +84,17 @@ export class GenericWebhooksRouter {
...
+                        sandbox: ''
+                    }
+                },
+                frameguard: { action: 'deny'},
```
davidegirardi, line 95:
```suggestion
                xFrameOptions: { action: 'deny'},
```

Does the same thing but hints better at what header it sets.

davidegirardi (changes requested):

PR #936, opened 2024-05-16
Fix GitLab's ready for review hook

zecakeh:
The code resulted on this error when there was a title change (rest of the stack trace omitted):

```
[Bridge] Connection GitLabRepo http://gitlab.local/group/website failed to handle gitlab.merge_request.update: TypeError: Cannot read properties of undefined (reading 'startsWith')
  at GitLabRepoConnection.onMergeRequestUpdate (~/Projets/matrix/hookshot/src/Connections/GitlabRepo.ts:630:53)
```

As far back as I could go in the docs (GitLab 14.10), the keys for changes have been `previous` and `current`. Besides, there is a `draft` change that we can use directly instead of guessing from the title.

I removed the `[key: string]` type declaration because the type of the value depends on the key of the change.

### DIFF
--- a/changelog.d/936.bugfix
+++ b/changelog.d/936.bugfix
@@ -1,0 +1,1 @@
+Fix GitLab's ready for review hook.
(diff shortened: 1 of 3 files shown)


Half-Shot (changes requested):
If you can do a changelog entry for this one, it'll be good to go :+1: . Thanks!

zecakeh:
Woops, I had forgotten to push the commit :sweat_smile:

PR #1013, opened 2025-02-07
Helm Chart - Fix missing namespace attributes on resources

Arkaniad:
I noticed that Hookshot doesn't want to install into specific namespaces under unique circumstances (in ArgoCD w/ Helmfile specifically) and it seems like one resource got the right config, but others didn't. This MR fixes the remainder.


Half-Shot (approved):
Looks good to me. Would you be able to do a changelog entry to describe the change?

Arkaniad:
Certainly!

Arkaniad:
Added changelog and checked merge when ready - thanks!

PR #1016, opened 2025-02-15
docs: explain how to access room state for gitlab projects

### LINKED ISSUE #260: Documentation nits: I try to add hookshot to a room.
I found a confusion here: 

https://matrix-org.github.io/matrix-hookshot/latest/usage/auth.html#github

I think a line inside github informing us how to use the oauth alternative in the DM room as an alternative to the PAT would be useful here.

I found this message is slightly confusing:
```
Failed to handle command: You must be able to set state in a room ('Change settings') in order to setup new integrations.
```

I think it would help if it says both you and I need this access, or perhaps "@michaelk:element.io and @hookshot:matrix.org" to be explicit about who needs what.

I found this to be helpful https://matrix-org.github.io/matrix-hookshot/latest/usage/room_configuration/github_repo.html

But I think this could be clearer that both the user running !hookshot and the hookshot bot both need permissions to configure things in the room.

It also says `This connection supports a few options which can be defined in the room state:`

Which i think is a requirement to head into /devtools and manually edit the room state. Being clearer about the format of this would be good.

It might be good to be explicit that "repo" and "org" are options too; but they should not be updated.

eg:
```
        "ignoreHooks": ["issue", "pull_request", "release"]
```

(Also, i'm not clear whether ignoring "issue" is enough to ignore all types under issue, or if we need to ignore the individual types explicitly)

The event type is also still:
  "type": "uk.half-shot.matrix-hookshot.github.repository",

Which is fine but maybe should be a org.matrix now the repository has changed hands.

gwennlbh:
Closes #260

### DIFF
--- a/docs/usage/room_configuration/gitlab_project.md
+++ b/docs/usage/room_configuration/gitlab_project.md
@@ -19,7 +19,7 @@ To set up a connection to a GitLab project in a new room:
 
 ## Configuration
 
-This connection supports a few options which can be defined in the room state:
+This connection supports a few options which can be defined in the room state[^2]:
 
 | Option | Description | Allowed values | Default |
 |--------|-------------|----------------|---------|
@@ -33,6 +33,8 @@ This connection supports a few options which can be defined in the room state:
 
 
 [^1]: `ignoreHooks` is no longer accepted for new state events. Use `enableHooks` to explicitly state all events you want to see.
+
+[^2]: To change room state, use Element on Desktop, open devtools with `/devtools` and click on "Explore room state". Next, choose `uk.half-shot.matrix-hookshot.gitlab.repository`. You'll then be able to see and edit (if you have the required permissions) the settings by setting properties on the JSON object directly.
 
 
 ### Supported event types


docs/usage/room_configuration/gitlab_project.md
```diff
@@ -34,6 +34,8 @@ This connection supports a few options which can be defined in the room state:
 
 [^1]: `ignoreHooks` is no longer accepted for new state events. Use `enableHooks` to explicitly state all events you want to see.
 
+[^2]: To change room state, use Element on Desktop, open devtools by typing `/devtools` + enter in the chat box, and click on "Explore room state". Next, choose `uk.half-shot.matrix-hookshot.gitlab.repository`. You'll then be able to see and edit (if you have the required permissions) the settings by setting properties on the JSON object directly.
```
Half-Shot, line 37:
```suggestion
[^2]: To change room state with Element Web/Desktop: open devtools by typing `/devtools` + enter in the chat box, and click on "Explore room state". Next, choose `uk.half-shot.matrix-hookshot.gitlab.repository`. You'll then be able to see and edit (if you have the required permissions) the settings by setting properties on the JSON object directly.
```

Half-Shot (commented):
Suggesting we also mention Web, since it's the same :).

Could you also complete a signoff https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/CONTRIBUTING.md#sign-off, and ideally a changelog entry too. (Sorry, I know this is a bit annoying since you are just making a drive-by change)

PR #1065, opened 2025-06-10
Add a docker-compose setup for easy setup for developers to hack on

Half-Shot:
This is a quick and dirty docker-compose setup (soon with accompanying docs) to get users going for the purposes of hacking on Hookshot. This is NOT intended for production and would be dangerous to run. This should give developers a complete setup to run Hookshot against a Synapse/Element pair.


docker-compose.yml
```diff
@@ -1,44 +1,59 @@
-version: '3.7'
+name: matrix-hookshot
```
Half-Shot, line 1: N.B. OpenProject were promised documentation on how this works :)

PR #1097, opened 2025-10-21
Add optional argument to "feed" command for specifying template

sysvinit:
This change adds a new optional argument to the "feed" command for specifying a template to be used for formatting feed entries. Specifying a template is otherwise only possible with widgets enabled and not through the command interface.

I've also seen that `string-argv` is used for parsing chat commands, which should correctly decode quoted arguments containing spaces, so I've added a note about this to the template placeholder documentation.

### DIFF
--- a/changelog.d/1097.feature.md
+++ b/changelog.d/1097.feature.md
@@ -1,0 +1,1 @@
+Add an optional template argument to the `feed` command, to allow specifying a template for feed entries when using the command interface.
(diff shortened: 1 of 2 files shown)


Half-Shot (approved):
This is *fine*, albeit this doesn't give a user a way to edit their template later. We probably need another command for that at some point.

**EDIT:** Can you also sign off the changes, and add a changelog entry please :)

PR #1106, opened 2025-11-23
Add support for 'relates_to' in GenericHook

gamedolphin:
This pull request makes a minor enhancement to the `GenericHookConnection` class, adding support for including a new property in the message content if present.

* Added support for the `relates_to` property in the outgoing message content, allowing messages to include relation information when available.

### DIFF
--- a/src/Connections/GenericHook.ts
+++ b/src/Connections/GenericHook.ts
@@ -623,7 +623,7 @@ export class GenericHookConnection
     }
 
     if (typeof safeData?.relates_to  === "object") {
-      msg.relates_to = safeData.relates_to;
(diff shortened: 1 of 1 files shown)


src/Connections/GenericHook.ts
```diff
@@ -620,6 +621,10 @@ export class GenericHookConnection
         msg.html = `<strong>${safeData.username}</strong>: ${msg.html}`;
       }
     }
+
+    if (typeof safeData?.relates_to  === "object") {
+      msg.relates_to = safeData.relates_to ?? undefined;
```
Half-Shot: I think this needs stronger validation, at least checking whether the object looks like a relates_to object?

PR #1111, opened 2025-12-08
Create format-as-code.js

HarHarLinks:
This snippet will receive a webhook containing `text`, e.g. Slack Webhook format and dump it into a codeblock.
This is useful for hooks which contain monospace-formatted info, for example logs or ZFS event daemon (https://openzfs.github.io/openzfs-docs/man/master/8/zed.8.html) notifications.

You might call this trivial, but I PR'd it anyway because sometimes it's just useful to get someone thinking, or to discover this accidentally when looking for "zed webhooks".

image


Half-Shot (approved):
Hah, I suppose that makes sense .

PR #1229, opened 2026-02-26
Ensure Generic webhooks can be statically configured with js transformation functions.

### LINKED ISSUE #1228: Static webhook setup with transformationFunction fails due to late quickjs initialisation
When using statically configured webhooks, a racecondition occurs when hookshot is started and the connection validator checks whether the quickjs engine has been loaded: https://github.com/matrix-org/matrix-hookshot/blob/main/src/generic/WebhookTransformer.ts#L45-L47

It seems that the engine is only initialized after the config has been loaded: https://github.com/matrix-org/matrix-hookshot/blob/main/src/App/BridgeApp.ts#L51-L53

Half-Shot:
Fixes #1228 

Not the prettiest fix, but just pass in the configured bridge settings when doing the initial validation.


wampiedriessen:
This PR seems to fix exactly what i'm getting stuck at, configuring hookshot.

Is there any way I can help get this resolved? I wouldn't mind testing this, if possible?

djohnson1865:
> This PR seems to fix exactly what i'm getting stuck at, configuring hookshot.
> 
> Is there any way I can help get this resolved? I wouldn't mind testing this, if possible?

I'm in the same boat. Ran into this issue a couple days after the PR written and have been waiting for months to test the fix in my setup. Seems like it's right at the finish line. Would be happy to assist in any way I can

Half-Shot:
Bleh, apologies this fell down the distraction hole. I'll do some updates today.

PR #1243, opened 2026-03-07
Fix incorrect usage of securityContext

qdii:
Fix a few issues:

*  `containerSecurityContext` does not exist in values.yaml or README.md
*  `securityContext` should be used for container-specific security contexts
*  `podSecurityContext` should be used for contexts that apply to all containers


regdude:
can someone please look into this PR?
Without this fix it is not possible to run hookshot in Kubernetes with CIS enabled.

PR #1257, opened 2026-04-07
feat: make webhook display name suffix configurable

ornariece:
### Summary

The suffix on a generic webhook user's displayname is hardcoded to ` (Webhook)`.
This adds a `generic.displaynameSuffix` option so it can be changed, or dropped
entirely by setting it to an empty string.

It defaults to ` (Webhook)`, so existing deployments are unaffected.

### Motivation

When a room only receives webhooks from one source, the connection name usually
already identifies the sender and the ` (Webhook)` suffix is just noise. It's most
noticeable in push notifications, where the sender name is shown prominently and
there isn't much room for it. Rooms fed by several sources can set a custom suffix
instead of removing it.

### Changes

- `generic.displaynameSuffix`, defaulting to ` (Webhook)`.
- `GenericHookConnection.ensureDisplayname()` uses it in place of the hardcoded
  string.
- Documented in `docs/setup/webhooks.md`; `config.sample.yml` regenerated.
- Tests for the config default, an empty suffix and a custom suffix, and for the
  displayname each produces.

### Example

```yaml
generic:
  enabled: true
  urlPrefix: https://example.com/webhook/
  userIdPrefix: _webhooks_
  displaynameSuffix: "" # connection name used as-is
```

The option only has an effect when `userIdPrefix` is set — without it, webhooks
post as the bridge bot and no displayname is set.


Half-Shot (approved):
Looks good to me 👍

Half-Shot (changes requested):
Hah, went too early. A few CI failures to clean up

ornariece:
@Half-Shot 
Rebased onto main and pushed a tidied-up version that addresses the branch-related CI failures:

- Added `changelog.d/1257.feature`.
- Regenerated `config.sample.yml`.
- Dropped `displaynameSuffix` from `GenericHookServiceConfig`. `GenericHookConnection`
  reads the bridge config directly, so the suffix never needed to go out over the
  widget API — adding it there is what broke the `publicConfig` assertions in
  `tests/config/Config.spec.ts`.
- Added tests for the config default, an empty suffix and a custom suffix, plus
  tests covering the displayname `ensureDisplayname()` actually sets. Documented
  the option in `docs/setup/webhooks.md`.

Run locally against current main, all green: `pnpm run lint:js`, `cargo fmt --check`,
`cargo clippy -- -Dwarnings`, the `config.sample.yml` diff check, the metrics-docs
diff check, `towncrier check`, `pnpm run build`, and `pnpm run test:cover` on both
Node 22 and Node 24.

`docker-build (amd64)` will likely still fail — it ends in a push to
`ghcr.io/matrix-org` and a fork PR's token can't write there
(`denied: installation not allowed to Write organization package`). Nothing this
branch can do about that one.

PR #1258, opened 2026-04-09
Fix error when handling replies with parents that can't be found

Half-Shot:
If a message contained a reply but hookshot could not read it, the message errors and is skipped. This change allows the handler to continue with a missing reply instead.

### DIFF
--- a/src/Bridge.ts
+++ b/src/Bridge.ts
@@ -1358,6 +1358,8 @@ export class Bridge {
 
     let replyEvent: RoomEvent<IJsonType> | undefined;
     if (event.content["m.relates_to"]?.["m.in_reply_to"]) {
+      const parentEventId =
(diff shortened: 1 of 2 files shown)


src/Bridge.ts
```diff
@@ -1373,11 +1375,16 @@ export class Bridge {
...
         }
-      } else {
-        // This is a new style reply.
-        const parentEventId =
-          event.content["m.relates_to"]?.["m.in_reply_to"].event_id;
-        replyEvent = await this.as.botClient.getEvent(roomId, parentEventId);
+      } else if (parentEventId) {
+        try {
+          // This is a new style reply.
+          await this.as.botClient.getEvent(roomId, parentEventId);
```
peace-maker, line 1381: This doesn't assign the event to `replyEvent`. Was this intentional?

Just a drive-by comment. I'm not familiar with the codebase.

PR #1303, opened 2026-06-08
Add ability to hint that URL previews should be off for messages.

Half-Shot:
Uses https://github.com/matrix-org/matrix-spec-proposals/pull/4095 to allow indicating that a message should not have any URL previews.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -273,7 +273,7 @@ if you have a configuration you'd like to apply every room.
 
 ```yaml
 messaging:
-    allowUrlPreviews: true
+  allowUrlPreviews: true
 ```
 
 ### Services configuration
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -266,6 +266,16 @@ cache:
   redisUri: "redis://redis-host:6379"
 ```
 
+### Messaging defaults
+
+Hookshot allows you to configure global overrides for certain message features. This is useful
+if you have a configuration you'd like to apply every room.
+
+```yaml
+messaging:
+    allowUrlPreviews: true
+```
+
 ### Services configuration
 
 You will need to configure some services. Each service has its own documentation file inside the setup subdirectory.
--- a/src/Connections/FeedConnection.ts
+++ b/src/Connections/FeedConnection.ts
@@ -11,7 +11,7 @@ import { readFeed, sanitizeHtml } from "../libRs";
 import UserAgent from "../UserAgent";
 import { retry, retryMatrixErrorFilter } from "../PromiseUtil";
 import { ConnectionType } from "./type";
-import { BridgeConfigMessaging } from "../config/sections/Messages";
+import { BridgeConfigMessaging } from "../config/sections";
 const log = new Logger("FeedConnection");
 const md = new markdown({
   html: true,
--- a/src/Connections/FeedConnection.ts
+++ b/src/Connections/FeedConnection.ts
@@ -11,6 +11,7 @@ import { readFeed, sanitizeHtml } from "../libRs";
 import UserAgent from "../UserAgent";
 import { retry, retryMatrixErrorFilter } from "../PromiseUtil";
 import { ConnectionType } from "./type";
+import { BridgeConfigMessaging } from "../config/sections/Messages";
 const log = new Logger("FeedConnection");
 const md = new markdown({
   html: true,
@@ -31,6 +32,7 @@ export interface FeedConnectionState extends IConnectionState {
   label: string | undefined;
   template: string | undefined;
   notifyOnFailure: boolean | undefined;
+  showUrlPreviews: boolean | undefined;
 }
 
 export interface FeedConnectionSecrets {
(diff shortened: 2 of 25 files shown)


src/config/sections/Messages.ts
```diff
@@ -0,0 +1,51 @@
...
+  public formatMatrixMessage<T extends Record<string, unknown>>(
+    eventContent: T,
+    messageConfig: BridgeConfigMessagingYAML = {},
+  ): MessageHints & T {
+    const hints: MessageHints = {};
+    if (!(messageConfig.allowUrlPreviews ?? this.allowUrlPreviews)) {
+      hints["com.beeper.linkpreviews"] = [];
+    }
+
+    console.log({ hints }, messageConfig, this.allowUrlPreviews);
```
AndrewFerr: Should this be `console.debug`?

src/config/sections/Messaging.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (!(messageConfig.allowUrlPreviews ?? this.allowUrlPreviews)) {
+      hints["com.beeper.linkpreviews"] = [];
+    }
+
+    console.log({ hints }, messageConfig, this.allowUrlPreviews);
+
+    return {
+      ...eventContent,
+      ...hints,
+    };
```
AndrewFerr, line 47: Instead of potentially injecting an empty `hints`, perhaps this could return `eventContent` unchanged when previews are allowed.

Something like:

```suggestion
    return !(messageConfig.allowUrlPreviews ?? this.allowUrlPreviews)
      ? {
          ...eventContent,
          ["com.beeper.linkpreviews"]: [],
        }
      : eventContent;
```

docs/setup.md
```diff
@@ -266,6 +266,16 @@ cache:
...
 ```
 
+### Messaging defaults
+
+Hookshot allows you to configure global overrides for certain message features. This is useful
+if you have a configuration you'd like to apply every room.
+
+```yaml
+messaging:
+    allowUrlPreviews: true
```
AndrewFerr: From CI
```suggestion
  allowUrlPreviews: true
```

src/Connections/FeedConnection.ts
```diff
@@ -31,6 +32,7 @@ export interface FeedConnectionState extends IConnectionState {
   label: string | undefined;
   template: string | undefined;
   notifyOnFailure: boolean | undefined;
+  showUrlPreviews: boolean | undefined;
```
AndrewFerr, line 35: Consider
```suggestion
  showUrlPreviews?: boolean | undefined;
```
which lets a test get away with not setting it.

src/Connections/FeedConnection.ts
```diff
@@ -224,6 +246,7 @@ export class FeedConnection extends BaseConnection implements IConnection {
     stateKey: string,
     private state: FeedConnectionState,
     private readonly intent: Intent,
+    private readonly msgConfig: BridgeConfigMessaging,
```
AndrewFerr, line 249: Consider
```suggestion
    private readonly msgConfig = new BridgeConfigMessaging(),
```
which lets a test get away with not setting it.

src/config/sections/Messages.ts
```diff
@@ -0,0 +1,51 @@
+import { ConfigError } from "../../Errors";
+import { configKey } from "../Decorators";
+
+interface BridgeConfigMessagingYAML {
+  allowUrlPreviews?: boolean;
+}
+
+interface MessageHints {
+  ["com.beeper.linkpreviews"]?: unknown[];
```
AndrewFerr: As long as this array is only ever set to be empty, consider typing it as just `[]` until support for non-empty elements is added.
```suggestion
  ["com.beeper.linkpreviews"]?: [];
```

src/config/sections/Messages.ts
AndrewFerr: This also triggers the `no-console` lint rule, so reminder to address that eventually

src/Connections/FeedConnection.ts
```diff
@@ -91,7 +99,7 @@ export class FeedConnection extends BaseConnection implements IConnection {
   }
 
   static validateState(
-    data: Record<string, unknown> = {},
+    data: Partial<Record<keyof FeedConnectionState, unknown>> = {},
```
AndrewFerr, line 102: This could be reverted because all callers pass a `Record<string, unknown>` here anyways.

AndrewFerr (approved):
Looks good, but have some comments/suggestions, with only the lint-satisfying ones being mandatory.

src/config/sections/Messages.ts
Half-Shot: Yeah this was left behind when debugging. I need to start catching these.

src/config/sections/Messaging.ts
Half-Shot: I think we're expecting to add more hints so will keep the present format atm :)

src/config/sections/Messages.ts
Half-Shot: Oh that's smart!

src/Connections/FeedConnection.ts
Half-Shot: Ah so this is me trying to be clever, the `keyof FeedConnectionState` just means my editor now know which keys to check. It's fundamentally the same as before but it's handy because it autocompletes in a nicer way.

PR #1323, opened 2026-06-11
Change changes_requested text to requested changes

### LINKED ISSUE #363: Change `changes_requested` text to `requested changes`
The bot posts a message saying `{user} changes_requested`, but this doesn't flow well grammatically.

image

I feel like it'd be better if it said something like this:

image

I thought about removing `in` to fit in with other stuff the bot says, but the above made the most sense to me.

adarsh3410:
This changes the GitHub review state text from
"changes_requested"
to
"requested changes"
to improve readability of review notifications.

Fixes #363

### DIFF
--- a/src/Connections/GithubRepo.ts
+++ b/src/Connections/GithubRepo.ts
@@ -1502,18 +1502,24 @@ export class GitHubRepoConnection
       throw Error("No repository content!");
     }
     const orgRepoName = event.repository.full_name;
-    const emojiForReview = {
+    const reviewState = event.review.state.toLowerCase();
+
+    const emojiForReviewMap: Record<string, string> = {
       approved: "✅",
-      // This apparently fires each time someone comments on the PR, which is not helpful.
-      //    'commented': '🗨️',
+      commented: "💬",
       changes_requested: "🔴",
-    }[event.review.state.toLowerCase()];
+    };
+
+    const emojiForReview = emojiForReviewMap[reviewState];
     if (!emojiForReview) {
       // We don't recongnise this state, run away!
       return;
     }
+
     const content = emojify(
-      `${emojiForReview} **${event.sender.login}** ${event.review.state.toLowerCase()} [${orgRepoName}#${event.pull_request.number}](${event.pull_request.html_url}) "${event.pull_request.title}"`,
+      `${emojiForReview} **${event.sender.login}** ${
+        reviewState === "changes_requested" ? "requested changes" : reviewState
+      } [${orgRepoName}#${event.pull_request.number}](${event.pull_request.html_url}) "${event.pull_request.title}"`,
     );
     await this.sendEvent(content, {
       ...FormatUtil.getPartialBodyForGithubIssue(


src/Connections/GithubRepo.ts
```diff
@@ -1502,19 +1502,32 @@ export class GitHubRepoConnection
       throw Error("No repository content!");
     }
     const orgRepoName = event.repository.full_name;
-    const emojiForReview = {
-      approved: "✅",
-      // This apparently fires each time someone comments on the PR, which is not helpful.
-      //    'commented': '🗨️',
```
Half-Shot, line 1508: I think we want to keep commented off for the time being due to the above comment

src/Connections/GithubRepo.ts
```diff
@@ -1502,19 +1502,32 @@ export class GitHubRepoConnection
...
+const reviewTextMap: Record<string, string> = {
+    approved: "approved",
+    commented: "commented",
+    changes_requested: "requested changes",
+};
+
+const reviewText = reviewTextMap[reviewState] || reviewState;
+
+  const content = emojify(
+ `${emojiForReview} **${event.sender.login}** ${reviewText} in ${orgRepoName}#${event.pull_request.number}`
```
Half-Shot: This seems to reformat the text output too:
`${emojiForReview} **${event.sender.login}** ${event.review.state.toLowerCase()} [${orgRepoName}#${event.pull_request.number}](${event.pull_request.html_url}) "${event.pull_request.title}"`

is now

`${emojiForReview} **${event.sender.login}** ${reviewText} in ${orgRepoName}#${event.pull_request.number}`

Can you revert the other changes but keep in reviewText?

Half-Shot (changes requested):
You will also need to sign off the changes please :)

src/Connections/GithubRepo.ts
adarsh3410: Thanks for the feedback! I've updated the PR to keep the existing formatting and only change the changes_requested text. I've also added the signoff.

adarsh3410:
Thanks for the feedback! I've updated the PR to keep the existing formatting and only change the `changes_requested` text to `requested changes`. I've also added the signoff and force-pushed the updated commit.

src/Connections/GithubRepo.ts
```diff
@@ -1502,19 +1502,31 @@ export class GitHubRepoConnection
...
     const orgRepoName = event.repository.full_name;
-    const emojiForReview = {
-      approved: "✅",
-      // This apparently fires each time someone comments on the PR, which is not helpful.
-      //    'commented': '🗨️',
-      changes_requested: "🔴",
-    }[event.review.state.toLowerCase()];
+   const reviewState = event.review.state.toLowerCase();
+
+const emojiForReviewMap: Record<string, string> = {
```
Half-Shot: Indent looks wrong to me, you probably need to run `pnpm run lint:js:apply`

src/Connections/GithubRepo.ts
```diff
@@ -1502,19 +1502,31 @@ export class GitHubRepoConnection
...
+
+
+
+
+  const content = emojify(
+ `${emojiForReview} **${event.sender.login}** ${
+     reviewState === "changes_requested"
+         ? "requested changes"
+         : reviewState
+ } ${orgRepoName}#${event.pull_request.number}`
```
Half-Shot: This is still missing the title and other parts

src/Connections/GithubRepo.ts
```diff
@@ -1502,19 +1502,31 @@ export class GitHubRepoConnection
...
     }
-    const content = emojify(
-      `${emojiForReview} **${event.sender.login}** ${event.review.state.toLowerCase()} [${orgRepoName}#${event.pull_request.number}](${event.pull_request.html_url}) "${event.pull_request.title}"`,
-    );
+   
+
+
+
+
+  const content = emojify(
```
Half-Shot: I don't think you need any of these 5 newlines.

Half-Shot (changes requested):

Issue #1353, opened 2026-06-24
OpenProject: getOpenProjectSafeUrl hardcodes http:// instead of preserving the original protocol, causing connection setup to fail on HTTPS instances


MacherelR:
### Describe the bug

When attempting to bridge an OpenProject project to a Matrix room using the `!hookshot openproject add https://your-op-instance/projects/<ID>` command, the command always fails with:

```
Failed to handle command: Expected 'url' to match the origin 'https://your-op-instance'
```

This happens on any OpenProject instance served over HTTPS.

---

### Root cause

In `src/Connections/SetupConnection.ts`, the `getOpenProjectSafeUrl` method builds the sanitised URL to pass to `provisionConnection` but hardcodes the `http://` scheme regardless of the original input:

```ts
// SetupConnection.ts — getOpenProjectSafeUrl
return `http://${url.host}/projects/${projectId}`;
```

This means a user input of `https://op.example.com/projects/42` produces the safe URL `http://op.example.com/projects/42`.

`provisionConnection` then calls `validateOpenProjectConnectionState`, which checks that the URL's origin matches `config.openProject.baseURL.origin`. Since the config stores `https://op.example.com` but the safe URL is `http://op.example.com`, the origins differ and an `ApiError` is thrown.

---

### Steps to reproduce

1. Deploy Hookshot with an OpenProject instance configured over HTTPS.
2. Invite the Hookshot bot to a Matrix room.
3. Run: `!hookshot openproject add https://your-op-instance/projects/<numeric-id>`
4. Observe the error: `Expected 'url' to match the origin 'https://your-op-instance'`

---

### Expected behaviour

The command succeeds and the room is bridged to the OpenProject project.

---

### Actual behaviour

The command fails with an origin mismatch error because the safe URL is built with `http://` instead of `https://`.

---

### Fix

Replace the hardcoded `http://` with the protocol taken from the input URL:

```ts
// Before
return `http://${url.host}/projects/${projectId}`;

// After
return `${url.protocol}//${url.host}/projects/${projectId}`;
```

---

### Environment

- Hookshot version: 7.3.2
- OpenProject instance: self-hosted, HTTPS
- Matrix homeserver: Synapse 1.155.0
- Deployment: Docker Compose

PR #1354, opened 2026-06-24
fix(openproject): preserve URL protocol in getOpenProjectSafeUrl

MacherelR:
Related to #1353 

### Problem

When a user runs `!hookshot openproject add https://op-instance/projects/<id>` against an OpenProject instance served over HTTPS, the command always fails with:

```
Failed to handle command: Expected 'url' to match the origin 'https://op-instance'
```

### Root cause

In `src/Connections/SetupConnection.ts`, `getOpenProjectSafeUrl` sanitises the user-provided URL and returns a normalised form to pass to `provisionConnection`. However, it hardcodes the `http://` scheme regardless of the input:

```ts
return `http://${url.host}/projects/${projectId}`;
```

So a user input of `https://op.example.com/projects/42` produces `http://op.example.com/projects/42`.

`provisionConnection` then calls `validateOpenProjectConnectionState`, which asserts:

```ts
if (parsedUrl.origin !== baseUrl.origin) {
    throw new ApiError(`Expected 'url' to match the origin '${baseUrl.origin}'`, ErrCode.BadValue);
}
```

Since `config.openProject.baseURL.origin` is `https://op.example.com` but the safe URL origin is `http://op.example.com`, the check always fails on HTTPS instances.

### Testing

- Deployed Hookshot 7.3.2 against a self-hosted OpenProject 17.5 instance over HTTPS
- Without the fix: `!hookshot openproject add https://op.example.com/projects/8` fails with origin mismatch
- With the fix: command succeeds and the room is bridged correctly

### DIFF
--- a/src/Connections/SetupConnection.ts
+++ b/src/Connections/SetupConnection.ts
@@ -874,7 +874,7 @@ export class SetupConnection extends CommandConnection {
         `The project url you entered was not valid. It should be part of ${expectedOrigin}`,
       );
     }
-    return `http://${url.host}/projects/${projectId}`;
+    return `${url.protocol}://${url.host}/projects/${projectId}`;
   }
 
   @botCommand("openproject add", {
--- a/src/Connections/SetupConnection.ts
+++ b/src/Connections/SetupConnection.ts
@@ -874,7 +874,7 @@ export class SetupConnection extends CommandConnection {
         `The project url you entered was not valid. It should be part of ${expectedOrigin}`,
       );
     }
-    return `${url.protocol}://${url.host}/projects/${projectId}`;
+    return new URL(`/projects/${projectId}`, url).toString();
   }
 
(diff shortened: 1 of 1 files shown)


src/Connections/SetupConnection.ts
```diff
@@ -874,7 +874,7 @@ export class SetupConnection extends CommandConnection {
         `The project url you entered was not valid. It should be part of ${expectedOrigin}`,
       );
     }
-    return `http://${url.host}/projects/${projectId}`;
+    return `${url.protocol}://${url.host}/projects/${projectId}`;
```
Half-Shot: This should probably just use
```suggestion
    return new URL(`/projects/${projectId}`, url).toString()
```

MacherelR: You're right, done :)

PR #1368, opened 2026-08-10
Update Cargo.lock

AndrewFerr:
Should have been done as part of 7.4.4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1101,7 +1101,7 @@ dependencies = [
 
 [[package]]
 name = "matrix-hookshot"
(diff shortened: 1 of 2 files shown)


AndrewFerr:
Forgot to commit this update to 7.4.4 despite [scripts/release.sh](https://github.com/matrix-org/matrix-hookshot/blob/main/scripts/release.sh).  That script should be more diligent & check that all updates are not just staged, but also committed.

Not sure if this needs a changelog, as IIRC nothing external consumes the Rust metadata anyways.

PR #1370, opened 2026-08-26
Update SDKs, update minimum Node version to 24

AndrewFerr:


AndrewFerr:
Note that the e2e failures are the same as what's been happening on `main` lately.  Example: https://github.com/matrix-org/matrix-hookshot/actions/runs/35734142183/job/106768129030

PR #1373, opened 2026-09-02
OpenProject module - Implement Phase 1: parity migration from PR#1064

rbondesson:

##Summary 
This is an initial implementation of a OpenProject module for element-web migrated from, https://github.com/matrix-org/matrix-hookshot/pull/1064. The PR gives a starting point for future imrpovements.

Signed-off-by: Rickard Bondesson <rickardb@element.io>


modules/openproject/element-web/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "@vector-im/compound-web": "^9.4.1",
+    "@types/node": "^22",
+    "@types/react": "^19",
+    "@vitejs/plugin-react": "^5.0.4",
+    "react": "^19.2.7",
+    "react-dom": "^19.2.7",
+    "rollup-plugin-external-globals": "^0.13.0",
+    "styled-components": "^6.1.18",
+    "typescript": "^5.7.2",
+    "vite": "^7.0.0"
```
Half-Shot: Wonder if we could make this use "catalog:" to share deps, as I think the main app also uses similar packages.

Half-Shot: If it's a faff, we can do this later.

rbondesson: My two cents. The OpenProject module is built in a separate repository/workspace, so it cannot consume Element Web’s catalog.  We can introduce a local catalog for Hookshot’s repo own duplicated build dependencies, but that would not share dependencies with Element Web, so I suggest keeping these ranges separate for now.

If we want true cross-repository sharing, it would require moving the module into Element Web’s monorepo somehow. Or do you have another solution in mind?

rbondesson: Added pnpm catalog for: @types/node, @vector-im/compound-design-tokens, @vector-im/compound-web, typescript and vite

Half-Shot (approved):
Very plausible. Let's go!

PR #1388, opened 2026-09-09
OpenProject module - Implement Phase 2: stabilize the implementation

rbondesson:

##Summary 
This PR aims to stabilize the implementation of a OpenProject module for element-web migrated from, https://github.com/matrix-org/matrix-hookshot/pull/1064. No new features added in this PR.

##Changes
- Updated the cache sequence in the bridge. Sometimes the new state was read before the old state resulting in 'no changes'
- Improved formatting and edge cases. E.g. percentage value of 0 is treated as 'no value'... which is not the same as an explicitly chosen 0 value
- Added focused unit tests for event contract, update handling and edge cases
- Hardened the renderer. Sanitize inputs instead of using dangerouslySetInnerHTML
- Added renderer tests
- Added a common storybook for the repo and use it for the renderer

Signed-off-by: Rickard Bondesson <rickardb@element.io>


pnpm-lock.yaml
Half-Shot, line 1: Goodness it went a bit mad in here :D

package.json
```diff
@@ -18,6 +18,7 @@
     "build:app": "tsc --project tsconfig.json",
     "build:app:rs": "napi build --dts ../src/libRs.d.ts --release --output-dir ./lib",
     "build:docs": "ts-node scripts/build-metrics-docs.ts > docs/metrics.md && mdbook build",
+    "build:storybook": "storybook build",
```
Half-Shot, line 21: Stupid question, what is this used for?

package.json
```diff
@@ -40,6 +41,8 @@
     "lint:types": "pnpm --filter openproject-plugin lint:types",
     "lint:rs": "cargo fmt --all -- --check && cargo clippy -- -Dwarnings",
     "lint:rs:apply": "cargo clippy --fix && cargo fmt --all",
+    "lint:storybook": "tsc --project tsconfig.storybook.json --noEmit",
+    "storybook": "storybook dev -p 6006",
```
Half-Shot: I think this is the default port anyway so probably overkill (not sure if it will automatically pick another port if in conflict with an existing storybook when the `-p` flag is given, which is useful)

.storybook/app-web-root.css
```diff
@@ -0,0 +1,37 @@
+/*
+ * Match the Element Web application shell without importing its component
+ * styles. This keeps stories focused on the component under test while still
+ * allowing their typography and root defaults to match the real application.
```
Half-Shot, line 4: Can you reference the file you got this from so we can cross reference later?

modules/openproject/element-web/src/components/WorkPackageActions.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { safeUrl } from "./validation";
+
+export function WorkPackageActions({ url }: { url: unknown }) {
+  return (
+    <Button
+      Icon={PopOutIcon}
+      as="a"
+      size="md"
+      kind="secondary"
+      href={safeUrl(url)}
```
Half-Shot: if unsafe, does this just do nothing? Should we hide the button?

modules/openproject/element-web/src/components/WorkPackageChangedDetails.tsx
```diff
@@ -0,0 +1,97 @@
...
+  workPackage,
+  changes,
+  descriptionAsDetails = false,
+}: {
+  workPackage: OpenProjectWorkPackageContent;
+  changes: OpenProjectWorkPackageChanges;
+  descriptionAsDetails?: boolean;
+}) {
+  let innerContent = null;
+  if (changes.assignee !== undefined) {
```
Half-Shot, line 20: I forget if multiple changes are possible?

modules/openproject/element-web/src/components/WorkPackageLink.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { safeUrl } from "./validation";
+
+export function WorkPackageLink({
+  url,
+  children,
+}: {
+  url: unknown;
+  children: React.ReactNode;
+}) {
+  return <a href={safeUrl(url)}>{children}</a>;
```
Half-Shot: ditto, if this is unsafe do we just end up with a broken link? Should this also be a target="_blank"?

modules/openproject/element-web/tests/fixtures/OpenProjectEventFixtures.ts
```diff
@@ -0,0 +1,108 @@
+/**
```
Half-Shot, line 1: I'd expect these tests to live within modules/openproject/element-web, and for us to setup a test runner config in there.

Half-Shot (changes requested):
Few things, mostly structural stuff

package.json
rbondesson: That was maybe added a little early. Its used in the netlify build in the next PR. I didn't include the workflow files in this one. I'd rather leave it as is

modules/openproject/element-web/src/components/WorkPackageChangedDetails.tsx
rbondesson: Yes, multi-changes are very much a thing. Basically all properties can change at the same time.
I have no idea why the the if-statement order is what it is. Its the same is in the original implementation.

.storybook/app-web-root.css
rbondesson: * As defined in the web shared components package, this is the root CSS for the Element Web application.
 * @link https://github.com/element-hq/element-web/blob/develop/packages/shared-components/.storybook/app-web-root.css

package.json
rbondesson: I'll move it to 9006 which seems out of the way of the docker stuff.

docker run \
    --name matrix-hookshot \
    -d \
    -p 9993:9993 \ # Homeserver port
    -p 9000:9000 \ # Webhook port
    -p 9002:9002 \ # Metrics port
    -v /etc/matrix-hookshot:/data \
    ghcr.io/matrix-org/matrix-hookshot:latest

.storybook/app-web-root.css
rbondesson: * As defined in the web shared components package, this is the root CSS for the Element Web application.
 * @link https://github.com/element-hq/element-web/blob/develop/packages/shared-components/.storybook/app-web-root.css

modules/openproject/element-web/src/components/WorkPackageActions.tsx
rbondesson: If added a check if the url is safe and return null if it isn't

modules/openproject/element-web/src/components/WorkPackageLink.tsx
rbondesson: I changed the package header link and the package title link to render a <span instead if the url is unsafe

modules/openproject/element-web/tests/fixtures/OpenProjectEventFixtures.ts
rbondesson: I agree. I split up the module tests and moved them into modules/openproject/element-web/tests. I also made sure the all module tests are run when the root pnpm test is called.

modules/openproject/element-web/src/components/WorkPackageChangedDetails.tsx
Half-Shot: Probably lazy coding on my part, if we can stack these up together if multiple changes happen then that would be nice :)

modules/openproject/element-web/src/components/WorkPackageStyles.ts
```diff
@@ -0,0 +1,63 @@
+import { styled } from "styled-components";
+
+export const WorkPackageMessage = styled.div`
```
Half-Shot, line 3: fwiw, wonder if we can use CSS modules over styled. Styled was a thing we had back when we didn't have support for modules (hence, hack for the old impl), but I think these days it's quite easy.

Half-Shot (approved):
Cool. I think the stacked changes would be nice but given the old impl didn't have it, we can leave it til later if you prefer.

I think the old code is misleading by using `styled` over css modules. If it's easy then modules would be preferred. If it's hard then we can do that as a followup.

modules/openproject/element-web/src/components/WorkPackageChangedDetails.tsx
rbondesson: Agreed. I'll implement it in a followup PR

modules/openproject/element-web/src/components/WorkPackageStyles.ts
rbondesson: Agreed. I'll implement it in a followup PR

PR #1402, opened 2026-09-21
OpenProject - Refactor the element-web module to MVVM

rbondesson:

## Summary
Refactor the open project module to the MVVM pattern using the view model helpers from element web shared components. 

## Changes
- Update the folder structure: models, utils, viewmodels
- Extract business logic from components into WorkPackageMessageViewModel
- Move business logic test into WorkPackageMessageViewModel.spec.ts

Signed-off-by: Rickard Bondesson [rickardb@element.io](mailto:rickardb@element.io)


.storybook/main.ts
```diff
@@ -13,11 +16,40 @@ const config = {
...
       },
       optimizeDeps: {
         // Compound Web imports this CommonJS package as an ES default.
         include: ["classnames"],
       },
+      // web-shared-components contains a few Node-compatible process checks.
+      // Storybook runs in the browser, so provide the same static replacement
+      // as the OpenProject module build.
+      define: {
+        "process.env.NODE_ENV": JSON.stringify("development"),
```
Half-Shot:
```
> JSON.stringify("development") === '"development"' 
true
```

```suggestion
        "process.env.NODE_ENV": '"development"',
```

vite.config.mjs
```diff
@@ -3,9 +3,15 @@ import preact from '@preact/preset-vite'
 import { resolve } from 'path'
 import alias from '@rollup/plugin-alias'
 
+// Storybook uses the React renderer, while the application uses Preact as its
```
Half-Shot, line 6: Fine for now, we probably just need to rip out preact at some point

modules/openproject/element-web/src/components/workPackage/DescriptionView.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export function DescriptionView({
+  plain,
+  html,
+}: {
+  plain: string;
+  html?: string;
+}) {
+  if (html) {
+    return <DescriptionText dangerouslySetInnerHTML={{ __html: html }} />;
```
Half-Shot: we have validated this? if so can we comment that we are okay with the danger

modules/openproject/element-web/src/components/workPackage/LinkView.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export function LinkView({
+  url,
+  children,
+}: {
+  url?: string;
+  children: React.ReactNode;
+}) {
+  if (url) {
+    return <a href={url}>{children}</a>;
```
Half-Shot: needs appropriate target and referrer properties

modules/openproject/element-web/src/viewmodels/workPackage/ChangeViewModel.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+  | { readonly kind: "subject" }
+  | { readonly kind: "type"; readonly currentName: string };
+
+/**
+ * Represents the same single changed-field precedence as the existing
+ * renderer. Rendering every changed field is intentionally a future
+ * behavioural change, not part of this refactor.
+ */
+export function createChangeViewModel(
```
Half-Shot: Sooooo, if I were to nitpick these are snapshots rather than view models. I would name these `createChangeSnapshot` etc. And then the view models can consume them.

Half-Shot (changes requested):

vite.config.mjs
rbondesson: Yeah, agree. And maybe add the widgets into storybook at the same time. It took a while to setup everything just to get screenshots of the config widget

modules/openproject/element-web/src/components/workPackage/LinkView.tsx
rbondesson: target="_blank" rel="noopener noreferrer"

modules/openproject/element-web/src/viewmodels/workPackage/ChangeViewModel.ts
rbondesson: Changed the following:
ChangeViewModel -> ChangeSnapshot
DescriptionViewModel -> DescriptionSnapshot
DetailsViewModel ->DetailsSnapshot
LabelViewModel -> LabelSnapshot

Kept:
CreatedMessageViewModel
UpdatedMessageViewModel

modules/openproject/element-web/src/components/workPackage/DescriptionView.tsx
rbondesson: **Note. Solutions replaced.**

I have added additional comments and I have changed the input to be:
`export function DescriptionView({ plain, html }: DescriptionSnapshot) {`
I also added better test coverage for the safeHtml handling

But if you mean if we validate using dangerouslySetInnerHTML at all, the answer is no.
The corresponding function is used in element-web HtmlUtils
```
/*
 * Given an untrusted HTML string, return a React node with an sanitized version
 * of that HTML.
 * @param insaneHtml - the input to sanitize
 * @param className - an optional class name to apply to the element
 * @param sanitizeParams - the params to use for sanitization
 */
export function sanitizedHtmlNode(
    insaneHtml: string,
    className?: string,
    sanitizeParams: HtmlSanitizeOptions = sanitizeHtmlParams,
): ReactNode {
    const saneHtml = sanitizeHtml(insaneHtml, sanitizeParams);

    return <div dangerouslySetInnerHTML={{ __html: saneHtml }} dir="auto" className={className} />;
}
```

rbondesson: New approach: Call safeUrl, safeHtml and safeColor directly from the Views.

